PR #56, opened 2021-09-30
add global api load wait and move blocks to features

codev0:
Fixes # .

- move block list to features
- add global api init

@reviewer-or-team


website/frontend/src/api/initApi.ts
```diff
@@ -12,3 +12,25 @@ export function initApi() {
...
+    this.address = address;
+  }
+
+  async init() {
+    // eslint-disable-next-line no-underscore-dangle
+    this._api = await GearApi.create({ providerAddress: this.address });
+  }
+}
+
+export const nodeApi = new NodeApi();
```
osipov-mit, line 36: Add an env var here

osipov-mit (changes requested):

website/frontend/src/api/initApi.ts
codev0: ok, could you tell me what kind of env variable?

osipov-mit (approved):

PR #80, opened 2021-10-12
add: copy program Id from list of programs

osipov-mit:


EugenWay:
Nice! Should we move copyToClickboar to helpers coz it repeats a lot?

PR #121, opened 2021-11-01
New backend logic

osipov-mit:
The transition from a monolithic architecture of backend to a microservices completed.

There are the following microservices:
- `Api gateway` allows to interact from the outside
- `Data storage` interact with DB that stores metadata and events that occurs in blockchain
- `Events listener` listens blockchain node and sends filtrered events to the kafka


ansible/tasks/nginx/proxy.conf
```diff
@@ -1,13 +1,13 @@
...
-    listen [::]:443 ssl;
-    listen 443 ssl; 
-    ssl_certificate /etc/nginx/cert/cert.crt;
-    ssl_certificate_key /etc/nginx/cert/private.key;
-    ssl_dhparam /etc/nginx/cert/dhparam.pem;        
+    # listen [::]:443 ssl;
+    # listen 443 ssl; 
+    # ssl_certificate /etc/nginx/cert/cert.crt;
+    # ssl_certificate_key /etc/nginx/cert/private.key;
+    # ssl_dhparam /etc/nginx/cert/dhparam.pem;
```
shamilsan, line 10: Why SSL has been disabled in Nginx?

osipov-mit: It has to be deleted. I don't know why this file was here

shamilsan (approved):
Huge changes :-)
I'd host Docker images in https://github.com/orgs/gear-tech/packages instead of Docker Hub but it is not quite important.

osipov-mit:
> Huge changes :-) I'd host Docker images in https://github.com/orgs/gear-tech/packages instead of Docker Hub but it is not quite important.

Yeah, It's a good idea, but a little later

aderrod94 (approved):

website/data-storage/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM node:16-alpine
```
codev0, line 1: There is good practice to specify docker image with hash.

> Be very specific! Instead of pulling a tag, pull an image using the specific SHA256 reference of the Docker image, which guarantees you get the same image for every pull. However notice that using a SHA256 reference can be risky, if the image changes that hash might not exist anymore.
https://snyk.io/blog/10-docker-image-security-best-practices/

website/data-storage/src/interfaces/metadata.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface Meta {
+  id: string;
+  program: string;
+  owner: string;
+}
+
+export interface AddMetaParams extends RequestParams {
+  programId: string;
+  signature: string;
+  meta: string | any;
```
codev0, line 12: i think `any` here is redundant

website/data-storage/src/messages/messages.service.ts
```diff
@@ -0,0 +1,127 @@
...
+    const where = {
+      chain: params.chain,
+      destination: params.destination,
+      genesis: params.genesis,
+      isRead: params.isRead,
+    };
+    console.log(where);
+    const [result, total] = await this.messageRepo.findAndCount({
+      where,
+      take: params.limit | 20,
```
codev0, line 78: this 20 and other bellow looks like the magic value

website/events-listener/src/index.js
```diff
@@ -0,0 +1,27 @@
...
+import { listen } from './events.js';
+import { KafkaProducer } from './producer.js';
+import { logger } from './logger.js';
+import config from './config.js';
+
+const log = logger('Main');
+
+const main = async () => {
+  log.info('App is running...');
+  console.log(config);
```
codev0, line 11: Logging config file is not secure, because it can contain sensitive information

website/events-listener/src/logger.js
```diff
@@ -0,0 +1,29 @@
+import winston from 'winston';
```
codev0, line 1: I've heard [pino](https://github.com/pinojs/pino) much faster than winston

website/frontend/src/api/initApi.ts
```diff
@@ -14,6 +14,8 @@ class NodeApi {
 
   private chain: string | null;
 
+  private genesis: any;
```
codev0, line 17:
```suggestion
  private genesis: string | null;
```

website/frontend/src/components/blocks/MetaForm/MetaForm.tsx
```diff
@@ -57,8 +58,11 @@ export const MetaForm: VFC<Props> = ({ programName, programHash }) => {
   const handleFilesUpload = useCallback(
     async (file) => {
       try {
-        const fileBuffer: any = await readFileAsync(file);
+        const fileBuffer: Buffer = (await readFileAsync(file)) as Buffer;
```
codev0, line 61:
```suggestion
        const fileBuffer = (await readFileAsync(file)) as Buffer;
```
adding type annotation is redundant, because type will be infer

codev0 (changes requested):

website/events-listener/src/logger.js
osipov-mit: I'll take a look at this

website/events-listener/src/index.js
osipov-mit: Yeah, I forgot to delete this

codev0 (approved):

EugenWay (approved):

PR #211, opened 2021-12-03
Wasm compiler service

osipov-mit:
- To compile rust project it necessary to make post request to `/wasm-compiler/build` endpoint with parameter `file` and its value have to be buffer from archived project in zip. This request returns id of compiling process.
- To get result of compiling process it necessary to make post request to `/wasm-compiler/get-wasm` endpoint with parameter `id` which have to be equal the value received on the previous step


website/wasm-compiler/src/util.ts
```diff
@@ -0,0 +1,31 @@
+const AdmZip = require('adm-zip');
+import { mkdirSync } from 'fs';
+import { join } from 'path';
+import config from './config/configuration';
+
+export function isWasm(fileName: string): boolean {
+  let ext = fileName.split('.');
+  return ext[ext.length - 1] === 'wasm';
```
codev0, line 8: I think its weak way to check file kind, maybe checking mime type is better?

website/wasm-compiler/src/util.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export function generateId(): string {
+  return Math.random().toString(36).substring(2);
```
codev0, line 12: wow, it's nice

osipov-mit: 😊

website/wasm-compiler/src/app.module.ts
```diff
@@ -0,0 +1,34 @@
...
+  imports: [
+    ConfigModule.forRoot({
+      load: [configuration],
+    }),
+    TypeOrmModule.forRootAsync({
+      imports: [ConfigModule],
+      useFactory: (configService: ConfigService) => ({
+        type: 'postgres',
+        host: configService.get('db.host'),
+        port: +configService.get<number>('db.port'),
```
codev0, line 19: I think casting to number via parseInt more reliable

codev0 (approved):

PR #218, opened 2021-12-19
Change gas limit default

aderrod94:
Changed gas limit default when we upload a program


osipov-mit:
I guess it's also necessary on the "send message" page

osipov-mit (changes requested):
I guess it's also necessary on the "send message" page

aderrod94:
> 



> I guess it's also necessary on the "send message" page

done)

osipov-mit (approved):

PR #220, opened 2021-12-19
update getting user

### LINKED ISSUE #182: Update balance automaticly
### Problem to Solve

Now if we get a test balance or spend gas, to show an updated balance is necessary to refresh the page.
Needs to update account balance automatically

### Possible Solution

Move account balance to global state and provide update balance action

### Notes

_No response_

codev0:
Resolves #182 

- update getting accounts
- refactor Modal component
- connect balance subscription
- add dapp extension dynamic import


EugenWay:
The problem is not solved. The extension does not connect randomly

EugenWay (approved):

PR #231, opened 2021-12-27
Add IndexedDB for the development node

aderrod94:
There's a new way to store programs when we use development node - IndexedDB, but in our case I've added a lib that make using indexedDB ease.


website/frontend/src/components/pages/Programs/children/Message/Message.tsx
```diff
@@ -53,16 +56,30 @@ export const Message: VFC<Props> = ({ programId, programName, handleClose }) =>
...
+      } else {
+        getMeta()
+          .then((res) => {
+            setMeta(JSON.parse(res.result.meta) ?? null);
+          })
+          .catch((err: RPCResponseError) => dispatch(AddAlert({ type: EventTypes.ERROR, message: err.message })))
+          .finally(() => {
+            setReady(true);
+          });
+      }
```
nikitayutanov, line 80:
```suggestion
const fetchMeta = useCallback(async (id: string) => {
    const apiRequest = new ServerRPCRequestService();

    return apiRequest.callRPC<GetMetaResponse>(RPC_METHODS.GET_METADATA, { id });
  }, []);

  const getMeta = chain === 'Development' ? localPrograms.getItem : fetchMeta;

  useEffect(() => {
    if (!meta) {
      getMeta(programId)
        .then((res) => {
          setMeta(JSON.parse(res.result.meta) ?? null);
        })
        .catch((err: RPCResponseError) => dispatch(AddAlert({ type: EventTypes.ERROR, message: err.message })))
        .finally(() => {
          setReady(true);
        });
    }
  }, [meta, programId, getMeta, dispatch]);
```
I suggest to think about one general variable that would choose what function you need, and then handle it's execution the same way regardless of its choice, though localPrograms.getItem should return the same structure as fetchMeta.
I think it's much clearer solution that probably can be applied to other components too,

website/frontend/src/helpers.ts
```diff
@@ -57,3 +60,13 @@ export const signPayload = async (injector: any, address: string, payload: any,
     console.error(err);
   }
 };
+
+export const getAllLocalPrograms = () => {
```
nikitayutanov, line 64:
```suggestion
export const getLocalPrograms = () => {
```

website/frontend/src/components/pages/Programs/children/Message/Message.tsx
```diff
@@ -19,6 +20,8 @@ type Props = {
 };
 
 export const Message: VFC<Props> = ({ programId, programName, handleClose }) => {
+  const chain = localStorage.getItem('chain');
```
nikitayutanov, line 23: `'chain'` should be in `consts.ts`

website/frontend/src/components/pages/Programs/children/Message/Message.tsx
```diff
@@ -53,16 +56,30 @@ export const Message: VFC<Props> = ({ programId, programName, handleClose }) =>
...
     if (!meta) {
-      getMeta()
-        .then((res) => {
-          setMeta(JSON.parse(res.result.meta) ?? null);
-        })
-        .catch((err: RPCResponseError) => dispatch(AddAlert({ type: EventTypes.ERROR, message: err.message })))
-        .finally(() => {
-          setReady(true);
-        });
+      if (chain === 'Development') {
```
nikitayutanov, line 59: `'Development'` to `consts.ts`

website/frontend/src/helpers.ts
```diff
@@ -57,3 +60,13 @@ export const signPayload = async (injector: any, address: string, payload: any,
     console.error(err);
   }
 };
+
```
nikitayutanov, line 63:
```suggestion
const isDevChain = () => localStorage.getItem('chain') === 'Development';
```
Add isDevelopmentChain() function to reuse it across components instead of `const chain = localStorage.getItem('chain');` and `if (chain === 'Development')`

website/frontend/src/components/pages/Programs/children/Message/Message.tsx
aderrod94: Yes, it's a very good idea 👍 but as for me the method **getItem** should returns the structure like our usual Program structure. However, If we really want to change it, guess, it's another task.

website/frontend/src/helpers.ts
```diff
@@ -57,3 +61,19 @@ export const signPayload = async (injector: any, address: string, payload: any,
...
+  return localPrograms
+    .iterate((elem: LocalProgramModel) => {
+      programs.push(elem);
+    })
+    .then(() => programs);
+};
+
+export const getMetaFromLocalProgram = (id: string) => {
+  return localPrograms.getItem(id).then((response: any) => JSON.parse(response.meta.meta));
```
codev0, line 76: try to implement it without any

codev0 (changes requested):

website/frontend/src/services/IndexedDB.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  public getStore(db: IDBDatabase) {
+    const tx = db.transaction(this.storeName, 'readwrite');
+    const store = tx.objectStore(this.storeName);
+
+    return store;
+  }
+
+  public add(db: IDBDatabase, data: any) {
```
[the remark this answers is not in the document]
aderrod94: oh no, this file is unnecessary and deleted)

codev0 (approved):

website/frontend/src/services/ApiService.tsx
```diff
@@ -157,25 +180,48 @@ export const addMetadata = async (
...
-        dispatch(AddAlert({ type: EventTypes.SUCCESS, message: `Metadata added successfully` }));
+    if (isDevChain()) {
+      localPrograms
+        .getItem(programId)
+        .then((res: any) => {
+          const newData = {
+            ...res,
+            meta: {
+              meta: JSON.stringify(meta),
+              metaFile: metaFile,
```
nikitayutanov, line 191:
```suggestion
              metaFile,
```

nikitayutanov (approved):

EugenWay (approved):

PR #246, opened 2021-12-29
Rework `ExecutionError` type

xhjkl:
This PR changes the JSONRPC response to nest error message in field `error.message`:
```
$ curl localhost:3000/api --data-binary ...
...
* upload completely sent off: 222 out of 222 bytes
< HTTP/1.1 201 Created
...
< 
* Connection #0 to host localhost left intact
{   "jsonrpc": "2.0",
    "id": "from-curl",
    "error": { "message": "it works" }
}
```

### DIFF
--- a/website/api-gateway/src/json-rpc/handler.ts
+++ b/website/api-gateway/src/json-rpc/handler.ts
@@ -78,18 +78,21 @@ export class RpcMessageHandler {
 
   executeMethod(method: (params: any) => Observable<any>, procedure: IRpcRequest) {
     const result = method(procedure.params);
-    if (result) {
-      return new Promise((resolve, reject) => {
-        result.forEach((value) => {
+    if (!result) {
+      return;
+    }
+    return new Promise((resolve) => {
+      result
+        .forEach((value) => {
           if (!value) {
             resolve(this.getResponse(procedure, { error: 'Service is not available' }));
-          } else if (value.error) {
-            resolve(this.getResponse(procedure, value));
+          } else if ('error' in value) {
+            resolve(this.getResponse(procedure, value.error));
           } else {
             resolve(this.getResponse(procedure, null, value));
           }
-        });
-      });
-    }
+        })
+        .finally(() => console.log('end'));
+    });
   }
 }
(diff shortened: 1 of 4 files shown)


website/api-gateway/src/json-rpc/handler.ts
```diff
@@ -78,18 +78,21 @@ export class RpcMessageHandler {
...
+          } else if ('error' in value) {
+            resolve(this.getResponse(procedure, value.error));
           } else {
             resolve(this.getResponse(procedure, null, value));
           }
-        });
-      });
-    }
+        })
+        .finally(() => console.log('end'));
```
osipov-mit, line 95: I guess it's unnecessary here

website/data-storage/src/consumer/consumer.service.ts
```diff
@@ -88,9 +88,9 @@ export class ConsumerService {
 
   async programData(params: FindProgramParams): Result<Program> {
     try {
-      return (await this.programService.findProgram(params)) || { error: new ProgramNotFound().message };
-    } catch (error) {
-      return { error: error.message };
+      return (await this.programService.findProgram(params)) || { error: { message: new ProgramNotFound().message } };
```
osipov-mit, line 91: I think it's better to move throwing this error to `findProgram` method

osipov-mit (changes requested):

website/api-gateway/src/json-rpc/handler.ts
xhjkl: True; I somehow thought it was here before me.

website/data-storage/src/consumer/consumer.service.ts
xhjkl: Sure; thanks.

website/data-storage/src/programs/programs.service.ts
```diff
@@ -84,7 +85,7 @@ export class ProgramsService {
       return program;
     } catch (error) {
       logger.error(error);
-      return null;
+      throw new ProgramNotFound(error.message);
```
osipov-mit, line 88: And one more thing
It's necessary to handle this exception in the places where this method is called. (setStatus, isInDb, addProgramInfo methods and a few more)

osipov-mit (changes requested):

website/data-storage/src/programs/programs.service.ts
osipov-mit: @xhjkl What about this?

PR #247, opened 2021-12-29
Rewrite `events-listener` in Typescript

xhjkl:
This PR adds loose preliminary typings to `events-listener` service.


website/events-listener/package.json
```diff
@@ -1,23 +1,24 @@
...
+  "version": "0.2.2",
   "description": "",
-  "main": "src/indes.js",
-  "type": "module",
+  "main": "src/indes.ts",
   "scripts": {
-    "start": "node dist/index.js",
-    "start:dev": "nodemon src/index.js",
-    "build": "cp -r src dist"
+    "start": "ts-node src/index.ts",
```
osipov-mit, line 7: I think it's better to use ts-node in `start:dev` script and add here `build` script with running `tsc` command and use `node build/index.js` in start script.

website/events-listener/src/config.ts
```diff
@@ -2,9 +2,9 @@ import dotenv from 'dotenv';
 import { strict as assert } from 'assert';
 dotenv.config();
 
-const checkEnv = (env) => {
+const checkEnv = (env: unknown): string => {
   assert.notStrictEqual(env, undefined);
-  return env;
+  return env as never;
```
osipov-mit, line 7: Why it's never?

osipov-mit (changes requested):

website/events-listener/src/config.ts
xhjkl: It's not really a never, just my my attempt to get away with casting to any ;3
Will think of better typings.

osipov-mit:
Are you sure that without specifying `outDir: ./build` in tsconfig.json compiled code will be in `build` directory?

website/events-listener/src/config.ts
osipov-mit: I guess `env` can only be a string

website/events-listener/package.json
```diff
@@ -2,22 +2,26 @@
...
   "description": "",
-  "main": "src/indes.js",
+  "main": "src/index.ts",
   "type": "module",
   "scripts": {
-    "start": "node dist/index.js",
-    "start:dev": "nodemon src/index.js",
-    "build": "cp -r src dist"
+    "start": "node --experimental-specifier-resolution=node dist/index.js",
+    "start:dev": "nodemon src/index.ts",
```
osipov-mit, line 9: You have to use ts-node with ts files. Nodemon will not work

osipov-mit (changes requested):

website/events-listener/package.json
osipov-mit: waiting its resolving
@xhjkl

xhjkl: I'm having another experience:
```
$ yarn start:dev
[nodemon] 2.0.15
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: ts,json
[nodemon] starting `ts-node src/index.ts`
08:14:18 INFO [Main] App is running...

```

osipov-mit: Cool

osipov-mit (approved):

PR #253, opened 2022-01-06
Rich search

xhjkl:
After these changes, one could search a program not only by its ID, but also by a substring of any of its fields.


osipov-mit (changes requested):
It doesn't work
I suppose you need to change request from `program.data` to `program.all`

aderrod94 (approved):

osipov-mit (approved):

PR #257, opened 2022-01-11
contracts: remove parameter `chain`

xhjkl:
Parameter `chain` removed from everywhere as redundant.

### DIFF
--- a/website/data-storage/src/programs/programs.service.ts
+++ b/website/data-storage/src/programs/programs.service.ts
@@ -28,13 +28,12 @@ export class ProgramsService {
 
   async addProgramInfo(
     id: string,
-    chain: string,
     genesis: string,
     name?: string,
     title?: string,
     meta?: Meta,
   ): Promise<Program> {
-    const program = await this.findProgram({ id, chain, genesis });
+    const program = await this.findProgram({ id, genesis });
     program.name = name;
     program.title = title;
     program.meta = meta;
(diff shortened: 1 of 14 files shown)


website/data-storage/src/programs/programs.service.ts
```diff
@@ -75,8 +74,8 @@ export class ProgramsService {
   }
 
   async findProgram(params: FindProgramParams): Promise<Program> {
-    const { id, chain, genesis, owner } = params;
-    const where = owner ? { id, chain, genesis, owner } : { id, chain, genesis };
+    const { id, genesis, owner } = params;
+    const where = owner != null ? { id, genesis, owner } : { id, genesis };
```
osipov-mit, line 78: Why is `!= null` necessary?

website/data-storage/src/programs/programs.service.ts
```diff
@@ -88,11 +87,11 @@ export class ProgramsService {
     }
   }
 
-  async setStatus(id: string, chain: string, genesis: string, status: InitStatus): Promise<Program> {
+  async setStatus(id: string, genesis: string, status: InitStatus): Promise<Program> {
     return new Promise((resolve) => {
       setTimeout(async () => {
-        console.log(id, chain, genesis, status);
-        let program = await this.findProgram({ id, chain, genesis });
+        console.log(id, genesis, status);
```
osipov-mit, line 93: This string is unnecessary, we can delete it

osipov-mit (changes requested):
Also you can delete `chain` from events-listener service. And from all entities in data-storage service.

website/data-storage/src/programs/programs.service.ts
xhjkl: Done.

osipov-mit (approved):

PR #264, opened 2022-01-14
Update getGasSpent rpc call

osipov-mit:


shamilsan (approved):
Checked the API side only, looks correct 👍

EugenWay (approved):

PR #279, opened 2022-01-19
dm data storage tests

xhjkl:
- data-storage: `let` -> `const`
- data-storage: `main.spec` -> `save.spec`
- data-storage: stricter env checks
- !wip


website/data-storage/src/save.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+      mockReset();
+    }
+  });
+
+  it('should record uploaded program', async () => {
+    // Given an initialized Programs Service,
+    // when:
+    await programsService.save({
+      // This number spells "test" if you squint hard enough.
+      id: 7357,
```
osipov-mit, line 71: Actually id is a string in the form of hex

xhjkl: Understood; now it is in tests too.

PR #280, opened 2022-01-20
CLI tool to work with `.meta.wasm` files

osipov-mit:
- Add cli tool to work with .meta.wasm files
[README](https://github.com/gear-tech/gear-js/blob/do-gear-meta-cli/utils/meta-cli/README.md)


utils/meta-cli/test/test.js
```diff
@@ -0,0 +1,50 @@
...
+    expect(await exec('decode 0x50494e47 -t String')).toBeDefined();
+  });
+
+  test('encode PING', async () => {
+    expect(await exec('encode PING -t String')).toBeDefined();
+  });
+
+  test('encode init_input', async () => {
+    expect(
+      await exec(`encode '{"amount": 8, "currency": "GRT"}' -t init_input -m test/wasm/demo_meta.meta.wasm`),
```
xrdavies, line 35: Could you add a more meaningful check on the result instead of `toBeDefined`? Since we are not sure if the result is correct or not when introducing modifications.

osipov-mit: This tool use @gear-js/api to encode and decode payloads and for everything else. So there is no need to check right results here, because it is checked on @gear-js/api. And this lib can't affect it

osipov-mit: I check here only commands execution

xrdavies: > This tool use @gear-js/api to encode and decode payloads and for everything else. So there is no need to check right results here, because it is checked on @gear-js/api. And this lib can't affect it

Understand. Thanks.

EugenWay:
Great job! A really useful utility.

EugenWay (approved):

shamilsan (approved):
Awesome 💟

xhjkl (approved):

PR #289, opened 2022-01-24
Refactor signAndSend method

osipov-mit:
Since `0.11.0` version of @gear-js/api method `signAndSend` will return object of type `Event` like in events subscriptions.

Example how to get messageId with program uploading:
```javascript
api.program.signAndSend(account, ({ events = [], status }) => {
  if (status.isFinalized) {
    console.log(status.asFinalized); // it returns hash of block, where transaction was added
  }
  events.forEach(({ event: { method, section, data } }) => {
      if (method === 'InitMessageDispatched') {
          console.log(data[0]['messageId'].toHuman());
      }
    });
});
```

@aderrod94 @EugenWay @codev0 @nikitayutanov

### DIFF
--- a/api/src/interfaces/transaction.ts
+++ b/api/src/interfaces/transaction.ts
@@ -1,5 +1,8 @@
 import { AddressOrPair, SignerOptions } from '@polkadot/api/types';
+import { Event } from '@polkadot/types/interfaces';
+import { ISubmittableResult } from '@polkadot/types/types';
+// export interface TransactionStatusCb {
+//   (event: Event): void | Promise<void>;
+// }
 
-export interface TransactionStatusCb {
-  (data: any): void | Promise<void>;
-}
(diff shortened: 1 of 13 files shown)


api/src/interfaces/transaction.ts
```diff
@@ -1,5 +1,8 @@
 import { AddressOrPair, SignerOptions } from '@polkadot/api/types';
+import { Event } from '@polkadot/types/interfaces';
+import { ISubmittableResult } from '@polkadot/types/types';
+// export interface TransactionStatusCb {
+//   (event: Event): void | Promise<void>;
+// }
```
nikitayutanov, line 6: Should it remain?

nikitayutanov (approved):

api/src/interfaces/transaction.ts
osipov-mit: Yeah, thank you. This file is unnecessary

PR #290, opened 2022-01-25
Save message processing errors

osipov-mit:
- Listen to `MessageDispatched` events
- Save errors
- Delete old backend sources
- Return `type: module` in events-listener package.json
- Add ts-node-dev to `start:dev` script in events-listener


website/events-listener/tsconfig.json
```diff
@@ -3,9 +3,9 @@
     "esModuleInterop": true,
     "moduleResolution": "node",
     "strict": true,
-    "target": "ESNext",
+    "target": "esNext",
```
xhjkl, line 6: I'm not against it, but genuinely — why? 🤔

xhjkl (approved):

website/events-listener/package.json
```diff
@@ -3,29 +3,30 @@
   "version": "0.1.2",
   "description": "",
   "main": "src/index.ts",
-  "//type": "module",
+  "type": "module",
   "scripts": {
     "start": "node --experimental-specifier-resolution=node dist/index.js",
-    "start:dev": "nodemon src/index.ts",
+    "start:dev": "ts-node-dev src/index.ts",
```
xhjkl, line 9: Good thinking! I somehow fixated that we need precisely `nodemon` package and nothing but it for some reason 😊

website/events-listener/tsconfig.json
osipov-mit: I did this accidentally. Just was trying different options

website/events-listener/package.json
osipov-mit: Yeah. There was a little trouble. When we don't use `type: module` in package.json the app breaks in `start` script. Otherwise the app breaks in `start:dev` script if we use nodemon

xhjkl (approved):

nikitayutanov (approved):

PR #295, opened 2022-01-27
Single package for backend interfaces

osipov-mit:
- Saving blockHash for every event
- Field timestamp for program and message entity
- Package with interfaces that are used by `api-gateway`, `data-storage`, `events-listener`
- Fix Dockerfiles


website/data-storage/src/consumer/consumer.controller.ts
```diff
@@ -9,57 +21,56 @@ export class ConsumerController {
   constructor(private readonly consumerService: ConsumerService) {}
 
   @MessagePattern('events')
-  async addEvent(@Payload() payload) {
-    logger.log(payload.value, 'AddEvent');
-    const genesis = payload.headers.genesis;
+  async addEvent(@Payload() payload: AddEventKafkaPayload<Keys, any>) {
+    // const genesis = payload.headers.genesis;
```
xhjkl, line 25: Stale code?

xhjkl (approved):

website/data-storage/src/consumer/consumer.controller.ts
osipov-mit: yeah, thank you

website/interfaces/src/kafka-payload.ts
```diff
@@ -0,0 +1,5 @@
+export interface KafkaPayload<T> {
+  headers: any;
```
codev0, line 2: if it is http headers, we could describe its type as `Record<string, string>`

nikitayutanov (approved):

website/interfaces/src/kafka-payload.ts
osipov-mit: No, it isn't

PR #304, opened 2022-01-28
Search "My Programs" as well

xhjkl:
Now a user could see their programs filtered in the same manner as in "All Programs".

### DIFF
--- a/website/data-storage/src/programs/programs.service.ts
+++ b/website/data-storage/src/programs/programs.service.ts
@@ -60,7 +60,6 @@ export class ProgramsService {
         { genesis, owner, id: likeTerm },
         { genesis, owner, title: likeTerm },
         { genesis, owner, name: likeTerm },
-        { genesis, owner, title: likeTerm },
       ],
       ...getPaginationParams(params),
       order: {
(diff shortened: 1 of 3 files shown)


website/data-storage/src/programs/programs.service.ts
```diff
@@ -53,8 +53,15 @@ export class ProgramsService {
...
   async getAllUserPrograms(params: GetAllProgramsParams): Promise<GetAllProgramsResult> {
+    const { genesis, owner, term } = params;
+    const likeTerm = term != null ? ILike(`%${escapeSqlLike(term)}%`) : void null;
     const [result, total] = await this.programRepo.findAndCount({
-      where: { owner: params.owner, genesis: params.genesis },
+      where: [
+        { genesis, owner, id: likeTerm },
+        { genesis, owner, title: likeTerm },
+        { genesis, owner, name: likeTerm },
+        { genesis, owner, title: likeTerm },
```
osipov-mit, line 63: I think this row unnecessary

osipov-mit (changes requested):

website/data-storage/src/programs/programs.service.ts
xhjkl: Removed.

osipov-mit (approved):

PR #315, opened 2022-01-31
Get block timestamp

osipov-mit:
- Add method to get timestamp of particular block
- One method to get block by blockHash or blockNumber
- Bump version to `0.11.3`

### DIFF
--- a/api/src/Blocks.ts
+++ b/api/src/Blocks.ts
@@ -1,10 +1,11 @@
 import { GearApi } from './GearApi';
+import { GetBlockError } from './errors/blocks.errors';
 import { AnyTuple, AnyNumber } from '@polkadot/types/types';
-import { Compact, GenericExtrinsic, Vec } from '@polkadot/types';
+import { u64, Compact, GenericExtrinsic, Vec } from '@polkadot/types';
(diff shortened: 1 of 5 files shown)


api/src/Blocks.ts
```diff
@@ -43,6 +66,20 @@ export class GearBlock {
...
+  /**
+   * Get timestamp of block
+   * @param hashOrNumber hash or number of particular block
+   * @returns
+   */
+  async getBlockTimestamp(hashOrNumber: `0x${string}` | Uint8Array | number): Promise<Compact<u64>> {
+    const block = await this.get(hashOrNumber);
+    const tsAsU8a = block.block.extrinsics.filter(
+      (value) => value.method.method === 'set' && value.method.section === 'timestamp',
+    )[0].data;
```
nikitayutanov, line 78:
```suggestion
    const tsAsU8a = block.block.extrinsics.find(
      (value) => value.method.method === 'set' && value.method.section === 'timestamp',
    ).data;
```
Is there a need to use `filter`? `find` should be faster

nikitayutanov (approved):

PR #317, opened 2022-02-02
Block timestamp is now taken from the chain regardless of the servers timezone

xhjkl:
Number taken from logs.

```
geartech=# select to_timestamp('1644074020.000')::timestamp;
    to_timestamp     
---------------------
 2022-02-05 15:13:40
(1 row)

geartech=# select timestamp from program order by timestamp asc;
      timestamp      
---------------------
...
 2022-02-05 15:13:40

```

### DIFF
--- a/website/api-gateway/package.json
+++ b/website/api-gateway/package.json
@@ -21,7 +21,7 @@
     "test:e2e": "jest --config ./test/jest-e2e.json"
   },
   "dependencies": {
-    "@gear-js/interfaces": "1.0.0",
+    "@gear-js/interfaces": "workspace:^",
     "@nestjs/common": "^8.1.2",
     "@nestjs/config": "^1.0.3",
     "@nestjs/core": "^8.1.2",
--- a/website/api-gateway/package.json
+++ b/website/api-gateway/package.json
@@ -21,7 +21,7 @@
     "test:e2e": "jest --config ./test/jest-e2e.json"
   },
   "dependencies": {
-    "@gear-js/interfaces": "workspace:^",
+    "@gear-js/interfaces": "1.0.0",
     "@nestjs/common": "^8.1.2",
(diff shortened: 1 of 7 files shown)


website/api-gateway/package.json
```diff
@@ -21,7 +21,7 @@
     "test:e2e": "jest --config ./test/jest-e2e.json"
   },
   "dependencies": {
-    "@gear-js/interfaces": "1.0.0",
+    "@gear-js/interfaces": "workspace:^",
```
osipov-mit, line 24: I guess it's better to leave this dependency in the packages' dependencies. Because they may use different versions at different times.

website/events-listener/src/events.ts
```diff
@@ -21,100 +22,122 @@ import { logger } from './logger';
...
 
     events.forEach(async ({ event: { data, method } }: any) => {
-      let eventData:
-        | InitMessageEnqueuedData
-        | DispatchMessageEnqueuedData
-        | LogData
-        | InitSuccessData
-        | InitFailureData
-        | MessageDispatchedData;
+      base.timestamp = (await api.blocks.getBlockTimestamp(blockHash)).toNumber();
```
osipov-mit, line 137: Why didn't you add this value where `base` var was declared?

osipov-mit (changes requested):

website/api-gateway/package.json
xhjkl: Point taken; versions changed back.

website/events-listener/src/events.ts
xhjkl: A silly oversight of me, now fixed. Thanks.

osipov-mit (approved):

osipov-mit (approved):

PR #319, opened 2022-02-03
Add 404 page

aderrod94:
- Add page with 404 error


xhjkl (approved):

website/frontend/src/components/pages/Message/Message.tsx
```diff
@@ -123,8 +123,10 @@ export const Message: FC = () => {
...
     </div>
   ) : (
-    <div className="message">
-      <Loader color="#fff" className="animation-rotate" />
-    </div>
+    <p>
+      <div className="message">
+        <Loader color="#fff" className="animation-rotate" />
+      </div>
+    </p>
```
nikitayutanov, line 130: should be `Spinner`

EugenWay (approved):

website/frontend/src/components/pages/Message/Message.tsx
```diff
@@ -123,8 +119,10 @@ export const Message: FC = () => {
       </div>
     </div>
   ) : (
-    <div className="message">
-      <Loader color="#fff" className="animation-rotate" />
-    </div>
+    <p>
```
codev0, line 122: this is wrong, [you cannot wrap `div` by `p`.](https://stackoverflow.com/questions/8397852/why-cant-the-p-tag-contain-a-div-tag-inside-it)

codev0 (approved):

PR #320, opened 2022-02-03
make `blockHash` of `IBaseDBRecord` optional

xhjkl:
Some fields of `IBaseDBRecord` are actually nullable, judging by the schema. This PR addresses that discrepancy.

This PR is a prerequisite to #318.

### DIFF
--- a/website/interfaces/src/general.ts
+++ b/website/interfaces/src/general.ts
@@ -3,8 +3,8 @@ export interface IGenesis {
 }
 
 export interface IBaseDBRecord<Timestamp extends Date | number> extends IGenesis {
-  blockHash: string;
-  timestamp: Timestamp;
+  blockHash?: string;
+  timestamp?: Timestamp;
 }
 
 export interface ISignature {


osipov-mit:
It's unnecessary

xhjkl:
> It's unnecessary

If we don't do that, we won't be able to mark BaseEntity's fields as nullable in strict mode:

```
export abstract class BaseEntity implements IBaseDBRecord<Date> {

...

  @Column({ nullable: true })
  timestamp?: Date;
//         ^                   Type 'string | undefined' is not assignable to type 'string'.
}
```

xhjkl:
This enables #318.

osipov-mit:
> > It's unnecessary
> 
> 
> 
> If we don't do that, we won't be able to mark BaseEntity's fields as nullable in strict mode:
> 
> 
> 
> ```
> 
> export abstract class BaseEntity implements IBaseDBRecord<Date> {
> 
> 
> 
> ...
> 
> 
> 
>   @Column({ nullable: true })
> 
>   timestamp?: Date;
> 
> //         ^                   Type 'string | undefined' is not assignable to type 'string'.
> 
> }
> 
> ```
> 
> 
> 
> 

Got it. It is temporary decision. I just need to clean database, when it will be possible. And then we can remove nullable option

PR #343, opened 2022-02-18
Fix payload format on the message page

aderrod94:


website/frontend/src/components/pages/Message/Message.tsx
```diff
@@ -63,7 +63,7 @@ export const Message: FC = () => {
         decodedPayload = createType.create(type, message.payload);
       }
 
-      setMessagePayload(JSON.stringify(decodedPayload));
+      setMessagePayload(JSON.stringify(decodedPayload, null, 4));
```
nikitayutanov, line 66:
```suggestion
      setMessagePayload(getPreformattedText(decodedPayload));
```
We have helper function

aderrod94: thank you)

nikitayutanov (approved):

PR #379, opened 2022-02-23
Refactor backend execution errors

osipov-mit:


aderrod94 (approved):

website/data-storage/src/programs/programs.service.ts
```diff
@@ -35,19 +36,17 @@ export class ProgramsService {
     try {
       return await this.programRepo.save(program);
     } catch (error) {
-      errorLog.error(error, 29);
+      errorLog.error(error, 37);
```
codev0, line 39: looks like magic number

website/data-storage/src/programs/programs.service.ts
```diff
@@ -90,28 +89,21 @@ export class ProgramsService {
...
     return new Promise((resolve) => {
       setTimeout(async () => {
-        const program = await this.findProgram({ id, genesis });
-        if (program) {
+        try {
+          const program = await this.findProgram({ id, genesis });
           program.initStatus = status;
           resolve(await this.programRepo.save(program));
+        } catch (error) {
+          errorLog.error(error, 101);
```
codev0, line 104: looks like magic number

osipov-mit: it's a string number) but maybe you are right

codev0 (approved):

PR #382, opened 2022-02-23
Update CreateType

osipov-mit:
- Update CreateType class
- Add `createPayloadTypeStructure` function
---
From now next functions are deprecated
- `getTypeStructure`
- `parseHexTypes` - use decodeHexTypes instead it
- `encode` and `decode` - use `create` method
---
- Bump version to `0.13.1`


EugenWay (approved):

api/src/create-type/generate.ts
```diff
@@ -0,0 +1,50 @@
...
+      name,
+      value,
+      count,
+    };
+  }
+  return {
+    type,
+    name,
+    value,
+  };
```
nikitayutanov, line 16:
```suggestion
  const tree = { type, name, value };
  return type === 'Array' ? { ...tree, count } : tree;
```

api/src/create-type/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export function checkTypeAndPayload(type: string, payload: any): string {
+  if (payload === undefined) {
+    throw new CreateTypeError('Payload is not specified');
+  }
+  if (!type) {
+    return 'Bytes';
+  }
+  return type;
```
nikitayutanov, line 21:
```suggestion
  return type || 'Bytes';
```

nikitayutanov (approved):

PR #491, opened 2022-03-17
Add select

nikitayutanov:
Adds select with tests, improves few input tests.


website/frontend/src/common/components/Select/Select.module.scss
```diff
@@ -0,0 +1,26 @@
...
+
+.text {
+  margin-right: 16px;
+}
+
+.select {
+  flex: 1;
+  padding: 10px 45px 10px 20px;
+  color: $textColor;
+  background: #282828 url('./images/arrow.svg') center right 16px no-repeat;
```
codev0, line 20: you could inline this image by using [this tool](https://yoksel.github.io/url-encoder/), and reduce extra symbols from svg by using [this tool](https://jakearchibald.github.io/svgomg/)

website/frontend/src/common/components/Select/images/arrow.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="12" height="8" viewBox="0 0 12 8" fill="none" xmlns="http://www.w3.org/2000/svg">
```
codev0, line 1: `<svg width="12" height="8" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="m1.219.588 4.881 6 4.881-6" stroke="#fff"/></svg>` [from svgomg](https://jakearchibald.github.io/svgomg/)

aderrod94 (approved):

PR #499, opened 2022-03-21
Add textarea

nikitayutanov:
### Add:

- Textarea component with tests

### DIFF
--- a/website/frontend/src/common/components/Textarea/Textarea.tsx
+++ b/website/frontend/src/common/components/Textarea/Textarea.tsx
@@ -1,14 +1,19 @@
 import React, { TextareaHTMLAttributes } from 'react';
+import clsx from 'clsx';
+import styles from './Textarea.module.scss';
 
 interface Props extends TextareaHTMLAttributes<HTMLTextAreaElement> {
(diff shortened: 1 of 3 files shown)


website/frontend/src/common/components/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { TextareaHTMLAttributes } from 'react';
+import clsx from 'clsx';
+import styles from './Textarea.module.scss';
+
+interface Props extends TextareaHTMLAttributes<HTMLTextAreaElement> {
+  label?: string;
+}
+
+const Textarea = ({ label, className, rows = 5, ...attrs }: Props) => {
```
codev0, line 9: could we think combine input and text area to one component [like here](https://github.com/gear-tech/gear-js/pull/499)?

codev0 (approved):

website/frontend/src/common/components/Textarea/Textarea.tsx
nikitayutanov: Something to think off

PR #506, opened 2022-03-23
Fix custom type registration

osipov-mit:
### Fix: 
- Remove `customType` parameter from `GearApiOption` and replace it with `types` 

### Remove:
- `allEvents` variable from `GearApi` class. You can use `gearApi.query.system.events` instead
<hr>

## Bump @gear-js/api to 0.15.2

### DIFF
--- a/api/src/GearApi.ts
+++ b/api/src/GearApi.ts
@@ -31,7 +31,7 @@ export class GearApi extends ApiPromise {
 
   constructor(options?: GearApiOptions) {
     const { types, providerAddress, ...restOptions } = options;
-    const provider = remain?.provider || new WsProvider(providerAddress ?? 'ws://127.0.0.1:9944');
+    const provider = restOptions?.provider || new WsProvider(providerAddress ?? 'ws://127.0.0.1:9944');
     const defaultTypes = types ? { ...types, ...gearTypes } : gearTypes;
 
     super({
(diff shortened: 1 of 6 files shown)


api/src/GearApi.ts
```diff
@@ -35,18 +30,13 @@ export class GearApi extends ApiPromise {
...
   constructor(options?: GearApiOptions) {
-    const provider = options?.provider
-      ? options.provider
-      : new WsProvider(options?.providerAddress ?? 'ws://127.0.0.1:9944');
+    const { types, providerAddress, ...remain } = options;
+    const provider = remain?.provider || new WsProvider(providerAddress ?? 'ws://127.0.0.1:9944');
+    const defaultTypes = {
+      ...(types || {}),
+      ...gearTypes,
+    };
```
nikitayutanov, line 38:
```suggestion
    const defaultTypes = types ? { ...types, ...gearTypes } : gearTypes;
```
imo it's more straightforward this way

api/src/GearApi.ts
```diff
@@ -35,18 +30,13 @@ export class GearApi extends ApiPromise {
   public code: GearCode;
 
   constructor(options?: GearApiOptions) {
-    const provider = options?.provider
-      ? options.provider
-      : new WsProvider(options?.providerAddress ?? 'ws://127.0.0.1:9944');
+    const { types, providerAddress, ...remain } = options;
```
nikitayutanov, line 33:
```suggestion
    const { types, providerAddress, ...restOptions } = options;
```

nikitayutanov (approved):

PR #535, opened 2022-03-31
Use blocktimestamp in readState

osipov-mit:
### Add:
- Opportunity to use exec::timestamp in meta_state function. It is reached by getting last block timestamp immediately after receiving the program pages and passing it to `readState` function
- Test for it (thanks @LouiseMedova for wasms)
- Bump @polkadot/api to 7.14.3

---
## Bump @gear-js/api to 0.16.1


api/src/wasm/importObj.ts
```diff
@@ -18,7 +25,9 @@ export default (memory: WebAssembly.Memory, showDebug?: boolean, inputValue?: Ui
     },
     free: (_pages) => {},
     gr_block_height: () => {},
-    gr_block_timestamp: () => {},
+    gr_block_timestamp: () => {
+      return timestamp;
+    },
```
nikitayutanov, line 30:
```suggestion
    gr_block_timestamp: () => timestamp,
```

api/src/State.ts
```diff
@@ -43,7 +42,11 @@ export class GearProgramState extends GearStorage {
     if (!program) {
       throw new ReadStateError('Program is terminated');
     }
+
     const pages = await this.gPages(programId, program);
+    const curBlock = await this.api.blocks.getFinalizedHead();
+    const blockTimestamp = await this.api.blocks.getBlockTimestamp(curBlock.toHex());
```
nikitayutanov, line 48:
```suggestion
    const block = await this.api.blocks.getFinalizedHead();
    const blockTimestamp = await this.api.blocks.getBlockTimestamp(block.toHex());
```

api/src/Blocks.ts
```diff
@@ -96,7 +96,7 @@ export class GearBlock {
    */
   async getEvents(blockHash: `0x${string}` | Uint8Array): Promise<Vec<FrameSystemEventRecord>> {
     const apiAt = await this.api.at(blockHash);
-    return await apiAt.query.system.events();
+    return (await apiAt.query.system.events()) as Vec<FrameSystemEventRecord>;
```
nikitayutanov, line 99:
```suggestion
    return apiAt.query.system.events();
```
`return await` [is redundant](https://eslint.org/docs/rules/no-return-await).
Also is there a need to cast `Vec<FrameSystemEventRecord>` type manually? `apiAt.query.system.events()` already returns `Promise<Vec<FrameSystemEventRecord>>` as far as I see

nikitayutanov (approved):

PR #544, opened 2022-04-04
Update `mailbox.read` and add `small pages`

osipov-mit:
### Depends on https://github.com/gear-tech/gear/pull/810 and https://github.com/gear-tech/gear/pull/824

- Reduce page size in `readState` function
- Add new types
- Refactor reading mailbox
---
## Bump @gear-js/api to 0.17.0


EugenWay (approved):

api/src/Mailbox.ts
```diff
@@ -26,8 +26,25 @@ export class GearMailbox {
...
+      if (keys.length === 0) {
+        return [];
+      }
+      const keyPrefixes = this.api.query.gear['mailbox'].keyPrefix(accountId);
+      const keysPaged = await this.api.rpc.state.getKeysPaged(keyPrefixes, 1000, keyPrefixes);
+      const mailbox = (await this.api.rpc.state.queryStorageAt(keysPaged)) as Option<StoredMessage>[];
+      return mailbox.map((value, index) => {
+        return [
+          keys[index].toHuman() as [AccountId, Hex],
+          this.api.createType('GearCoreMessageStoredStoredMessage', value.unwrap()).toHuman() as any,
```
nikitayutanov, line 44:
```suggestion
      const mailbox = await this.api.query.gear.mailbox(accountId, messageId);
      return mailbox.toHuman() as IMailbox;
    } else {
      const keys = await this.api.query.gear.mailbox.keys(accountId);
      if (keys.length === 0) {
        return [];
      }
      const keyPrefixes = this.api.query.gear.mailbox.keyPrefix(accountId);
      const keysPaged = await this.api.rpc.state.getKeysPaged(keyPrefixes, 1000, keyPrefixes);
      const mailbox = (await this.api.rpc.state.queryStorageAt(keysPaged)) as Option<StoredMessage>[];
      return mailbox.map((option, index) => {
        return [
          keys[index].toHuman() as [AccountId, Hex],
          this.api.createType('GearCoreMessageStoredStoredMessage', option.unwrap()).toHuman() as unknown as HumanedMessage,
```
Still seems hacky as hell, but it's better than `any`

api/src/Mailbox.ts
```diff
@@ -41,20 +58,20 @@ export class GearMailbox {
...
-  unsubscribe() {
-    this.subscription.then((fn) => {
-      fn();
-    });
-  }
+  // unsubscribe() {
+  //   this.subscription.then((fn) => {
+  //     fn();
+  //   });
+  // }
```
nikitayutanov, line 76: Should this section remain?

osipov-mit: yes, i will return these methods later

nikitayutanov (approved):

PR #546, opened 2022-04-05
Add modal

nikitayutanov:
### Add:
- `Modal` component (yet only supports one open modal at a time)
- `transition` mixin
- `Button` has `type='button'` by default

### Fix:
- Peer dependencies
- Button style resets (`border`, `font-family`, `cursor`)

### DIFF
--- a/utils/gear-ui/src/components/Modal/Modal.module.scss
+++ b/utils/gear-ui/src/components/Modal/Modal.module.scss
@@ -30,6 +30,7 @@ $border: url("data:image/svg+xml,%3Csvg width='430' height='431' fill='none' xml
 }
 
 .heading {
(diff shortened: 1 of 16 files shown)


utils/gear-ui/src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  border-radius: $borderRadius;
+}
+
+.button {
+  margin-left: auto;
+  margin-bottom: 24px;
+}
+
+.heading {
+  // TOFIX: replace with common shared class/mixin when done
```
codev0, line 33: in my code editor highlighted words is `TODO` and `FIX`, i think in vscode the same

codev0 (approved):

PR #569, opened 2022-04-13
Fix error alerts, drop rpc getResource method

nikitayutanov:
Bumps deps and fixes error alerts messages display


website/frontend/src/services/ServerRPCRequestService.ts
```diff
@@ -24,18 +24,12 @@ type RPCRequest = {
...
-  data: string;
-};
-
 // eslint-disable-next-line @typescript-eslint/naming-convention
 type _RPCResponse<Result> = {
   id: JsonRpcId;
   jsonrpc: JsonRpcVersion;
   result: Result;
-  error: RPCError;
+  error: string;
```
codev0, line 32: Its not by specification https://www.jsonrpc.org/specification#response_object

nikitayutanov: Thanks, gonna be resolved on backend.

nikitayutanov: Fixed.

website/frontend/src/components/blocks/ProgramSwitch/ProgramSwitch.tsx
```diff
@@ -27,12 +27,12 @@ export const ProgramSwitch: VFC<Props> = ({ pageType }) => {
         throw new Error(`WALLET NOT CONNECTED`);
       }
 
-      const response = await apiRequest.getResource(RPC_METHODS.GET_TEST_BALANCE, {
+      const response = await apiRequest.callRPC(RPC_METHODS.GET_TEST_BALANCE, {
         address: `${currentAccount.address}`,
       });
 
       if (response.error) {
-        alert.error(`${response.error.error}`);
```
osipov-mit, line 35: Is there a way to reach to this line? If there will be an error in the response, `callRpc` method will throw an error, I suppose

website/frontend/src/components/blocks/ProgramSwitch/ProgramSwitch.tsx
```diff
@@ -27,12 +27,12 @@ export const ProgramSwitch: VFC<Props> = ({ pageType }) => {
...
+        alert.error(`${response.error.message}`);
```
osipov-mit, line 35: Is there a way to reach to this line? If there will be an error in the response, callRpc method will throw an error, I suppose

nikitayutanov: Alert gets called on error afterall, but I agree that it's a weird way to catch it and should be refactored.

osipov-mit (approved):

PR #588, opened 2022-04-18
Do js wrapper

osipov-mit:
Resolves # .

-
-
-

@reviewer-or-team


utils/js-wrapper/examples/guestbook/wrapper/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(TypeInfo, Decode, Encode, Serialize)]
+pub struct MessageIn {
+    pub author: String,
+    pub msg: String,
+}
+
+#[wasm_bindgen]
+pub fn __send_add_message(author: String, msg: String) -> Vec<u8> {
+    let mut v: Vec<u8> = Vec::new();
+    v.push(0);
```
breathx, line 20: This kind of pushing is discussable I think. I don't find that wrapper is enough comfortable to use and write manually

PR #611, opened 2022-04-25
feature: add name, version values on sidebar btn

Pinkdesu:
Resolves # .

- Display runtime name and version

### DIFF
--- a/website/frontend/src/components/blocks/Header/children/Menu/Menu.module.scss
+++ b/website/frontend/src/components/blocks/Header/children/Menu/Menu.module.scss
@@ -34,11 +34,11 @@
 }
 
 .link { 
+  @extend %menuItem;
+
   display: block;
   text-align: center;
   cursor: pointer;
-   
-  @extend %menuItem;
 
   &.active {
     box-shadow: inset 0 -3px 0 $successColor;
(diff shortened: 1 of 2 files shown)


website/frontend/src/components/blocks/Header/children/Menu/Menu.tsx
```diff
@@ -14,16 +14,26 @@ type ClassNameProps = {
...
+  const specName = api?.runtimeVersion.specName.toHuman();
+  const specVersion = api?.runtimeVersion.specVersion.toHuman();
+
   return (
     <ul className={styles.menu}>
       <li>
-        <span className={styles.link} onClick={openSidebar}>
-          {isApiReady ? localStorage.chain : 'Loading...'}
-        </span>
+        <button className={clsx(styles.sidebarBtn, !isApiReady && styles.loading)} onClick={openSidebar}>
```
aderrod94, line 27: We have @gear-js/ui with Button component. You should use it instead native button)

Pinkdesu: I think that it's wrong solution, because that button has a default styles and don't have 'children' prop.

aderrod94: yes, you're right 👍

osipov-mit (approved):

website/frontend/src/components/blocks/Header/children/Menu/Menu.module.scss
```diff
@@ -18,23 +18,63 @@
...
+      overflow-wrap: anywhere;
+
+      &.small {
+        font-size: $fontSizeSmall;
+      } 
+
+      &:not(:last-child) {
+         margin-bottom: 2px;
+      }
+   }
```
nikitayutanov, line 79: `overflow-wrap: anywhere` has this behaviour, I think we should drop it:
image
`text-align: right` seems to be more appropriate here and you can specify it in a parent

website/frontend/src/components/blocks/Header/children/Menu/Menu.tsx
```diff
@@ -14,16 +14,26 @@ type ClassNameProps = {
...
+          {isApiReady && (
+            <>
+              <span>{localStorage.chain}</span>
+              <span className={styles.small}>
```
nikitayutanov, line 31: `small` describes modificator, I believe w/ className we should describe element. For example: `runtime`

nikitayutanov (approved):

PR #614, opened 2022-04-27
Update upload metadata page

Pinkdesu:
- The upload metadata button is not accessible to outsiders
- Move the metadata update to a separate page
- Created a common page header
- Move the metadata form and fields to a separate catalogue
- Update the upload metadata page


website/frontend/src/components/blocks/legacy/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,47 @@
...
+    formId = 'meta-form';
+  }
+
+  return (
+    <div className="message-header">
+      <div className="message-header--info">
+        <button
+          type="button"
+          aria-label="arrowBack"
+          onClick={() => handleClose()}
```
codev0, line 32: no need to wrap it with arrow func

Pinkdesu: it's legacy component, I only moved it. It's will remove later.

website/frontend/src/components/blocks/MetaForm/UploadMetaForm/UploadMetaForm.module.scss
```diff
@@ -0,0 +1,41 @@
+@use '@gear-js/ui/breakpoints' as *;
+@use '@gear-js/ui/variables' as *;
+
+.uploadMetaForm {
+   display: 'flex';
+   flex-direction: 'column';
```
nikitayutanov, line 6:
```suggestion
   display: flex;
   flex-direction: column;
```

website/frontend/src/components/pages/Programs/children/All/All.tsx
```diff
@@ -79,35 +52,13 @@ export const All: VFC = () => {
...
-                  >
-                    <img src={UploadIcon} alt="upload-program" />
-                  </button>
-                </div>
-              </div>
-            );
-          })) ||
-          null}
+        {programs.map((program: ProgramModel) => (
+          <UserProgram key={program.id} program={program} disabledMeta={publicKeyRaw !== program.owner} />
```
nikitayutanov, line 56:
```suggestion
          <UserProgram key={program.id} program={program} />
```
I'd say no need for additional prop, since we're already have `program` prop and can do this check inside

website/frontend/src/components/blocks/MetaForm/UploadMetaForm/UploadMetaForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { readFileAsync } from 'helpers';
+import { useAccount } from 'hooks';
+import { addMetadata } from 'services/ApiService';
+
+type Props = {
+  programId: string;
+  programName: string;
+};
+
+const UploadMetaForm = ({ programId, programName }: Props) => {
```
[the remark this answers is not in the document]
Pinkdesu: I moved the common components to a separate folder. This is a separate feature, let it be here

website/frontend/src/components/pages/Programs/children/All/All.tsx
Pinkdesu: I don't think that's right. It's a simple component

nikitayutanov (approved):

osipov-mit (approved):

PR #621, opened 2022-04-29
WIP: add new meta parser based in new api

codev0:
- add meta parse based on new api


website/frontend/src/components/MetaFields/new-meta-parser.ts
```diff
@@ -0,0 +1,354 @@
...
+  value: MetaItemValue;
+  name: string;
+} | null;
+
+const NoneField = { None: { type: 'Primitive', name: 'None', value: 'None' } };
+
+function processFields(data: MetaItem, _path?: string[], isField = false): StackItem | null {
+  const path = _path || [];
+
+  if (data.type === 'Enum') {
```
Pinkdesu, line 95: You can use switch - case.

website/frontend/src/components/MetaFields/new-meta-parser.ts
```diff
@@ -0,0 +1,354 @@
...
+    payload.value.forEach((item) => {
+      if (isMetaItem(item)) {
+        stackRef.push(processFields(item, [...payload.path, '__fields']));
+        return;
+      }
+    });
+  } else {
+    Object.entries(payload.value)
+      .reverse()
+      .forEach((item) => {
```
Pinkdesu, line 185: If only values are needed, you can use Object.values

website/frontend/src/components/MetaFields/new-meta-parser.ts
```diff
@@ -0,0 +1,354 @@
...
+
+        setWith(
+          result.__values,
+          current.path.filter((i) => !['__root', '__fields'].includes(i)),
+          current.value === 'None' ? 'None' : '',
+          Object
+        );
+      } else if (isObject(current.value)) {
+        // region Parse if it is Struct
+        if (current.kind === 'Struct') {
```
Pinkdesu, line 241: switch - case

website/frontend/src/components/MetaFields/new-meta-parser.ts
```diff
@@ -0,0 +1,354 @@
...
+        //endregion
+      }
+    }
+  }
+
+  return result;
+}
+
+export function parseMeta(data: MetaItem): MetaFieldsStruct | null {
+  if (isObject(data)) {
```
Pinkdesu, line 342: isObject(data) && Object.keys(data).length !== 0

website/frontend/src/components/MetaFields/new-meta-parser.ts
```diff
@@ -0,0 +1,354 @@
...
+        // region Parse if it is Array
+        else if (current.kind === 'Array') {
+          setUpField(result, current.path, current.name);
+          processStackItem(stack, current);
+        }
+        // endregion
+        //region Parse if it is enum
+        else if (current.kind === 'Enum') {
+          setUpField(result, current.path, current.name, true);
+          Object.entries(current.value)
```
Pinkdesu, line 260: Object.values

website/frontend/src/components/MetaFields/new-meta-parser.ts
```diff
@@ -0,0 +1,354 @@
...
+  while (stack.length > 0) {
+    const current = stack.pop();
+
+    if (current) {
+      if (!result.__values) {
+        result.__values = {};
+      }
+
+      // Check if this primitive
+      if (current.kind === 'Primitive') {
```
Pinkdesu, line 220: you can make an enum describing 'kind'

website/frontend/src/components/MetaFields/new-meta-parser.ts
nikitayutanov: Agree, something like:
```suggestion
    switch (type) {
    case 'Enum':
    case 'Option':
    case 'Result':
      return {
        kind: data.type,
        path: [...path, data.name],
        value: data.value,
        name: data.name,
      }
    }
```

website/frontend/src/components/MetaFields/new-meta-parser.ts
codev0: i think it is not convenient and excess to use enum in case literal types

PR #644, opened 2022-05-16
Fix sort for the local programs

aderrod94:
- Fix sort local programs when we connected to the local node
- Fix error in **All** component. GearKeyring.decodeAddress parameter have to be **0x00** but not **''**

### DIFF
--- a/website/frontend/src/helpers.ts
+++ b/website/frontend/src/helpers.ts
@@ -96,7 +96,12 @@ export const getLocalPrograms = (params: any) => {
       }
     })
     .then(() => {
-      data.result.programs.sort((prev, next) => (prev.timestamp > next.timestamp ? -1 : 1));
+      data.result.programs.sort((prev, next) => {
+        const prevDateVal = new Date(prev.timestamp).valueOf();
+        const nextDateVal = new Date(next.timestamp).valueOf();
+
+        return prevDateVal > nextDateVal ? -1 : 1;
+      });
 
       return data;
     });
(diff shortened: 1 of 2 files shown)


Pinkdesu (approved):

website/frontend/src/helpers.ts
```diff
@@ -96,7 +96,12 @@ export const getLocalPrograms = (params: any) => {
       }
     })
     .then(() => {
-      data.result.programs.sort((prev, next) => (prev.timestamp > next.timestamp ? -1 : 1));
+      data.result.programs.sort((prev, next) => {
+        const prevDateVal = new Date(prev.timestamp).valueOf();
+        const nextDateVal = new Date(next.timestamp).valueOf();
+
+        return prevDateVal > nextDateVal ? -1 : 1;
+      });
```
nikitayutanov, line 104:
```suggestion
      data.result.programs.sort((prev, next) => prev.timestamp - next.timestamp);
```
Won't it work?

nikitayutanov: is it unix?

aderrod94: Снимок экрана 2022-05-16 в 13 21 02

nikitayutanov (approved):

website/frontend/src/helpers.ts
Pinkdesu: Optional, but you can use Date.parse() to not create classes

website/frontend/src/helpers.ts
```diff
@@ -96,7 +96,9 @@ export const getLocalPrograms = (params: any) => {
       }
     })
     .then(() => {
-      data.result.programs.sort((prev, next) => (prev.timestamp > next.timestamp ? -1 : 1));
+      data.result.programs.sort(
+        (prev, next) => new Date(next.timestamp).getTime() - new Date(prev.timestamp).getTime()
```
Pinkdesu, line 100: Optional, but you can use Date.parse to not create classes

PR #674, opened 2022-05-27
Add new payload form

Pinkdesu:
- Add new payload form (payload parser)
- Change payload from on pages Send, State, Upload program


osipov-mit (changes requested):
Pls check that the form's behavior is the same in all cases

osipov-mit (approved):
great job!

website/frontend/src/components/pages/State/State.tsx
```diff
@@ -76,6 +57,27 @@ const State: VFC = () => {
...
+  const typeStructures = useMemo(() => {
+    if (types && stateInput) {
+      const decodedTypes = decodeHexTypes(types);
+
+      return {
+        manual: createPayloadTypeStructure(stateInput, decodedTypes, true),
+        payload: createPayloadTypeStructure(stateInput, decodedTypes),
+      };
+    }
+  }, [types, stateInput]);
```
nikitayutanov, line 79: This part is the same on the few pages, so it can be standalone `getPayloadTypeStructures` function

PR #676, opened 2022-05-30
Add waitlist

osipov-mit:
## Changes
- Add GearWaitlist class to read programs' waitlists
- Add tests
- Update readme
---
## @gear-js/api version is `0.19.1`


nikitayutanov (approved):

codev0 (approved):

api/src/Waitlist.ts
```diff
@@ -22,10 +27,10 @@ export class GearWaitlist {
    * console.log(waitlist);
    * ```
    */
-  async read(programId: Hex, messageId?: Hex): Promise<WaitlistType> {
+  async read(programId: Hex, messageId?: Hex): Promise<WaitlistItem[] | WaitlistItem> {
```
nikitayutanov, line 30: Is there the ability to provide consistent return? I mean, `WaitlistItem[]` even for a single item.
It would be more convenient to deal with it on the frontend, if we will know that returned value is always an array.

osipov-mit: Override doesn't solve this problem?

osipov-mit: 15-16 rows

nikitayutanov: No, I meant like a simple check if there's a single `WaitlistItem` to still return it in an array with length 1

api/test/programs/Makefile
```diff
@@ -0,0 +1,15 @@
...
+
+cp-waitlist: 
+	cp test-waitlist/target/wasm32-unknown-unknown/release/test_waitlist.*.wasm ../wasm
+
+all: build-waitlist wasm-proc-waitlist cp-waitlist
+
+clear:
+	rm -rf ./**/target
+
+.PHONY: all clean
```
breathx, line 15: newline

api/test/programs/test-waitlist/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "test-waitlist"
+version = "0.1.0"
+authors = ["Gear Technologies"]
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
```
breathx, line 8:
```suggestion
```

As we discussed, you can now use `gear_wasm_builder`

api/test/programs/test-waitlist/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+        input: Vec<u8>,
+        output: Vec<u8>,
+}
+
+#[gstd::async_main]
+async fn main() {
+    let reply: Result<String, _> = msg::send_and_wait_for_reply(msg::source(), String::from("to waitlist"), 0)
+        .unwrap()
+        .await;
+}
```
breathx, line 17: it can be minimazed if need with just 
```rust
fn handle {
    gstd::exec::wait();
}
```

breathx (approved):

api/test/programs/test-waitlist/Cargo.toml
osipov-mit: Only in case of building through gear builder?

breathx: > Only in case of building through gear builder?

Yeah

api/test/programs/test-waitlist/.binpath
```diff
@@ -0,0 +1 @@
+target/wasm32-unknown-unknown/debug/test_waitlist.opt.wasm
```
breathx, line 1: this file may be gitignored

PR #686, opened 2022-06-01
disable upload button if balance not enough

codev0:
- disable upload button if balance not enough


website/frontend/src/utils/index.ts
```diff
@@ -0,0 +1 @@
+export * from './utils'
```
Pinkdesu, line 1: new line

Pinkdesu (approved):

PR #694, opened 2022-06-03
Update wasm-compiler service

osipov-mit:


nikitayutanov (approved):

website/wasm-compiler/src/server.ts
```diff
@@ -0,0 +1,60 @@
...
+export class Server {
+  dbService: DBService;
+  compiler: CompilerService;
+  app: Express;
+
+  constructor(dbService: DBService, compiler: CompilerService) {
+    this.dbService = dbService;
+    this.compiler = compiler;
+    this.app = express();
+    this.app.use(json());
```
codev0, line 20: it seems to need add default error handler for not caught errors, otherwise application crushes because of this

website/wasm-compiler/package.json
```diff
@@ -4,56 +4,39 @@
...
-    "@nestjs/platform-express": "8.4.5",
-    "@nestjs/typeorm": "8.0.4",
-    "@nestjs/websockets": "8.4.5",
     "adm-zip": "0.5.9",
-    "dockerode": "3.3.1",
-    "nats": "2.7.1",
-    "pg": "8.7.3",
+    "dotenv": "^16.0.1",
+    "express": "^4.18.1",
+    "multer": "^1.4.5-lts.1",
```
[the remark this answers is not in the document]
osipov-mit: I guess it isn't important here, but i'll take a look at this

website/wasm-compiler/src/server.ts
osipov-mit: There are only 2 endpoints in the service, so I guess try / catch is more than enough. And this service won't be used much

Pinkdesu (approved):

Almazatun (approved):

PR #718, opened 2022-06-09
Update events on backend

osipov-mit:
Update events-listener, data-storage and e2e-tests according to new events


website/data-storage/src/messages/messages.service.ts
```diff
@@ -121,27 +122,12 @@ export class MessagesService {
...
-    if (logMessages.length > 0) {
-      logMessages.forEach((log) => {
-        log.replyError = error;
-        this.messageRepo.save(log);
-      });
+  async setDispatchedStatus({ statuses, genesis }: IMessagesDispatchedKafkaValue): Promise<void> {
+    for (let messageId of Object.keys(statuses)) {
+      if (statuses[messageId] === 'Success') continue;
+
+      await sleep(100);
```
Almazatun, line 129: Simple refactor. Call two independent features.

```javascript
Promise.all([
sleep(10),
this.messageRepo.update({ genesis, id: messageId }, { processedWithPanic: true })
])
```

osipov-mit: No, these aren't independent. sleep() is necessary, because sometimes it happens that this method is called before the message that needs to be update is saved in database

Almazatun: Alright. I got it.

Almazatun (approved):

PR #722, opened 2022-06-13
Add @gear-js/common package for share data between services

Almazatun:
Resolves # .

All sharing data is contained in a `@gear-js/common` service.

@reviewer-or-team


package.json
```diff
@@ -3,8 +3,7 @@
   "private": true,
   "scripts": {
     "build:backend": "yarn build:interfaces && yarn build:jsonrpc-errors && yarn workspaces foreach -ptR --from '{@gear-js/event-listener,@gear-js/api-gateway,@gear-js/data-storage,@gear-js/test-balance}' run build",
```
osipov-mit, line 5:
```suggestion
    "build:backend": "yarn build:common && yarn workspaces foreach -ptR --from '{@gear-js/event-listener,@gear-js/api-gateway,@gear-js/data-storage,@gear-js/test-balance}' run build",
```

package.json
```diff
@@ -29,6 +28,7 @@
     "website/frontend",
     "website/test-balance",
     "website/interfaces",
+    "website/common",
     "website/jsonrpc-errors"
```
osipov-mit, line 32:
```suggestion
    "website/common"
```

website/events-listener/src/producer.ts
```diff
@@ -1,9 +1,8 @@
-import { Kafka, Producer } from 'kafkajs';
+import { logger } from '@gear-js/common';
 
-import { logger } from './logger';
+import { Kafka, Producer } from 'kafkajs';
 import config from './config';
-
-const log = logger('KafkaProducer');
+import { EVENTS_LISTENER } from './index';
```
osipov-mit, line 5: I suppose, there is a circular dependency here. KafkaProducer exported here is imported in index.ts and EVENTS_LISTENER exported in index is imported here.

website/events-listener/src/producer.ts
```diff
@@ -26,10 +25,11 @@ export class KafkaProducer {
     const admin = this.kafka.admin();
     try {
       await admin.connect();
-      log.info('Admin is connected');
+      logger.info(`${EVENTS_LISTENER} Kafka producer: Admin is connected`);
```
osipov-mit, line 28: EVENTS_LISTENER actually isn't necessary here. It's obvious. Kafka producer is more than enough

osipov-mit: The same thing in all cases of this file

website/events-listener/src/index.ts
```diff
@@ -1,17 +1,17 @@
...
 import { restartIfNeeded, setRestartNeeded } from './lifecycle';
 import { changeStatus } from './healthcheck';
 
-const log = logger('Main');
+export const EVENTS_LISTENER = 'EVENTS_LISTENER';
 
 const main = async () => {
   while (true) {
-    log.info('Starting...');
+    logger.info(`${EVENTS_LISTENER} Starting...`);
```
osipov-mit, line 14: May be it could just be `Main`?

website/data-storage/src/middleware/formResponse.ts
```diff
@@ -1,13 +1,13 @@
...
 
-const logger = new Logger('Middleware');
 export function FormResponse(target: Object, propertyKey: string, descriptor: TypedPropertyDescriptor<any>) {
   const originalMethod = descriptor.value;
   descriptor.value = async function SafeWrapper() {
     try {
       return { result: await originalMethod.apply(this, arguments) };
     } catch (ex) {
-      logger.warn(ex);
+      logger.warn(`${DATA_STORAGE} Middleware: ${JSON.stringify(ex)}`);
```
osipov-mit, line 10: It's obvious that it is data storage.

osipov-mit (changes requested):

website/test-balance/src/kafka.consumer.ts
```diff
@@ -1,9 +1,10 @@
+import { JSONRPC_ERRORS } from '@gear-js/common';
+
 import { Consumer, Kafka, KafkaMessage, Producer } from 'kafkajs';
 import config from './config';
 import { DbService } from './db';
 import { GearService } from './gear';
 import { KafkaLogger, Logger } from './logger';
-import errors from '@gear-js/jsonrpc-errors';
 
 const log = Logger('KafkaConsumer');
```
osipov-mit, line 9: Don't you want to replace logger here as well?

package.json
Almazatun: Fixed

package.json
Almazatun: Fixed

website/events-listener/src/producer.ts
Almazatun: It might be a not good idea. Without the `EVENTS_LISTENER` in `terminal` will little bit hurt to recognize from where the logger was called. For that, I include the service `name`.

Or do you @osipov-mit  think it will be better without `service name`?

website/events-listener/src/index.ts
Almazatun: @osipov-mit there is also the same thing. Try to imagine that you want to know every service in running `terminal`.

But in `terminal` you have only `Main` name it doesn't tell us what the service is running exactly.

website/events-listener/src/producer.ts
Almazatun: Moved `const` service name to `common` ----> `constant.ts` file.

image


Now looks like this. What do you think  @osipov-mit ?
```javascript 
import { EVENTS_LISTENER } from "./common/constant";
```

osipov-mit (approved):

PR #734, opened 2022-06-16
Remove manual payload

Pinkdesu:
- Remove manual payload
- Refactor metadata form
- Add the payload type field when there is no metadata


osipov-mit (approved):

website/frontend/src/components/blocks/UploadMeta/UploadMeta.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import styles from './UploadMeta.module.scss';
+import { UploadData } from './types';
+import { getMetaProperties } from './helpers';
+
+import { useAlert } from 'hooks';
+import { readFileAsync, checkFileFormat } from 'helpers';
+import { formStyles, FormText } from 'components/common/Form';
+
+type Props = {
```
aderrod94, line 13: I suggest to store **Props** in the local types.ts.

website/frontend/src/components/common/Form/FormPayloadType/FormPayloadType.tsx
```diff
@@ -3,14 +3,17 @@ import { useField } from 'formik';
 import clsx from 'clsx';
 import { Input, Checkbox } from '@gear-js/ui';
 
-import styles from './PayloadType.module.scss';
+import styles from '../Form.module.scss';
 
-import { formStyles } from 'components/common/Form';
+type Props = {
```
aderrod94, line 8: Also suggest to store Props in the local types.ts

website/frontend/src/components/common/Form/FormText/FormText.tsx
```diff
@@ -0,0 +1,25 @@
+import clsx from 'clsx';
+import { inputStyles, textareaStyles } from '@gear-js/ui';
+
+import styles from '../Form.module.scss';
+
+type Props = {
```
aderrod94, line 6: as I suggested above

website/frontend/src/components/blocks/UploadMeta/UploadMeta.tsx
nikitayutanov: > I suggest to store **Props** in the local types.ts.

why tho? we declare `Props` in the same component file everywhere, unless it's too complicated, and I see no reason to do otherwise

website/frontend/src/components/pages/Meta/children/UploadMetaForm/UploadMetaForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+};
+
+const UploadMetaForm = ({ programId, programName }: Props) => {
+  const alert = useAlert();
+  const { account } = useAccount();
+
+  const [meta, setMeta] = useState<Metadata | null>(null);
+  const [metaBuffer, setMetaBuffer] = useState<string | null>(null);
+  const [initialValues, setInitialValues] = useState<FormValues>({
+    name: programName,
```
aderrod94, line 26: use short form here { name }, If it possible

website/frontend/src/components/pages/Programs/children/Upload/children/UploadForm/UploadForm.tsx
```diff
@@ -1,94 +1,54 @@
...
-import { MetaSwitch } from 'components/common/MetaSwitch';
-import { META_FIELDS } from 'components/blocks/UploadMetaForm/model/const';
-import { FormInput, FormPayload, FormTextarea, FormNumberFormat } from 'components/common/Form';
-import { getMetaValues } from 'components/blocks/UploadMetaForm/helpers/getMetaValues';
+import { Fieldset } from 'components/common/Fieldset';
+import { FormInput, FormPayload, FormPayloadType, FormNumberFormat, formStyles } from 'components/common/Form';
+import { UploadMeta, UploadData } from 'components/blocks/UploadMeta';
 
 type Props = {
   setDroppedFile: Dispatch<SetStateAction<DroppedFile | null>>;
```
aderrod94, line 24: I suggest to store Props in the local types.ts.

website/frontend/src/components/blocks/UploadMeta/UploadMeta.tsx
aderrod94: When we create a new component and write some CSS. We can write only one rule but we store it separately. One type is the same thing and as for me better to store it separately too.

website/frontend/src/components/blocks/UploadMeta/UploadMeta.tsx
```diff
@@ -0,0 +1,84 @@
...
+  onReset: () => void;
+  onUpload: (data: UploadData) => void;
+};
+
+const UploadMeta = (props: Props) => {
```
aderrod94, line 18: I think better to get onReset, onUpload functions here

website/frontend/src/components/blocks/UploadMeta/UploadMeta.tsx
Pinkdesu: We have a different approach everywhere. And the example with css is generally incorrect, since it is a different language.

aderrod94 (approved):

PR #744, opened 2022-06-18
Refactor main page

Pinkdesu:
- Refactor all, recent, messages, upload pages
- Fix search logiс


website/frontend/src/components/pages/Programs/children/Messages/Messages.tsx
```diff
@@ -1,37 +1,47 @@
...
+      searchParams.set(URL_PARAMS.PAGE, String(1));
+      searchParams.set(URL_PARAMS.QUERY, '');
+      setSearchParams(searchParams);
+      setMessages([]);
+      setMessagesCount(0);
+    }
+
+    return () => {
+      isAccountLoaded.current = Boolean(account);
+    };
```
nikitayutanov, line 37: why not just looking on `account` itself?

website/frontend/src/components/pages/Programs/children/UserProgram/UserProgram.tsx
```diff
@@ -1,67 +1,72 @@
...
-import Copy from 'assets/images/copy.svg';
+import { Button } from '@gear-js/ui';
+
 import styles from './UserProgram.module.scss';
 
+import { routes } from 'routes';
+import { useAlert } from 'hooks';
+import { copyToClipboard, fileNameHandler, formatDate } from 'helpers';
+import { ProgramModel, ProgramStatus } from 'types/program';
+import copySGV from 'assets/images/copy.svg';
```
nikitayutanov, line 11:
```suggestion
import copySVG from 'assets/images/copy.svg';
```

website/frontend/src/components/pages/Programs/children/Messages/Messages.tsx
Pinkdesu: Required to track logout, because at the first render account is undefined

nikitayutanov: using useChangeEffect and looking for account is more straightforward imo

Pinkdesu: will not work, you need to skip the first two renders with an undefined account and the first time you got an account

PR #758, opened 2022-06-22
Update @gear-js/test-balance with new features @gear-js/api (transfer )

Almazatun:
Resolves # .

- Update `transferBalance` in @gear-js/test-balance
- Simple refactor @gear-js/test-balance
-

@reviewer-or-team

### DIFF
--- a/website/test-balance/src/domain/gear.service.ts
+++ b/website/test-balance/src/domain/gear.service.ts
@@ -1,11 +1,10 @@
-import { GearApi, GearKeyring } from '@gear-js/api';
+import { GearApi, GearKeyring, TransferData } from '@gear-js/api';
 import { initLogger } from '@gear-js/common';
-
-import config from '../config/config';
-
 import { BN } from '@polkadot/util';
 import { KeyringPair } from '@polkadot/keyring/types';
 import { DbService } from '../database/db';
+
+import config from '../config/config';
 
 const log = initLogger('GearService');
 
@@ -21,7 +20,7 @@ export class GearService {
     this.dbService = dbService;
   }
 
-  async connect() {
+  public async connect() {
     this.api = await GearApi.create({ providerAddress: config.gear.providerAddress });
     this.api.on('error', () => {
       GearApi.create({ providerAddress: config.gear.providerAddress }).then(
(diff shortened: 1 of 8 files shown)


website/test-balance/src/kafka/kafka.consumer.ts
```diff
@@ -61,10 +59,10 @@ export class KafkaConsumer {
       return { error: JSONRPC_ERRORS.InternalError.name };
     }
 
-    if (payload.genesis === this.gearService.genesisHash) {
+    if (payload.genesis === this.gearService.getGenesisHash) {
       try {
         if (await this.dbService.possibleToTransfer(payload.address, payload.genesis)) {
-          result = { result: await this.gearService.transferBalance(payload.address) };
+          result = { result: await this.gearService.transferBalance(payload.address, this.gearService.rootAccount) };
```
osipov-mit, line 65: no way
You should transfer balance not from the root account, but just from account.
Maybe it's better to make this.account as default in transfer balance method. And only in case of transfer to this.account use rootaccount. 
Root account is only needed to top up the account balance.
Maybe we should rename them

website/test-balance/src/domain/gear.service.ts
```diff
@@ -34,36 +33,64 @@ export class GearService {
...
+    return { status: 'ok', transferredBalance: balance.toString() };
   }
 
-  get genesisHash(): string {
+  public get getGenesisHash(): string {
     return this.api.genesisHash.toHex();
   }
 
-  async accountBalanceIsSmall(): Promise<boolean> {
+  private async transferData(from: KeyringPair): Promise<TransferData> {
```
osipov-mit, line 66: let's call it sendTransaction or smth like this. If we make it as a separate method

osipov-mit (changes requested):

website/test-balance/src/domain/gear.service.ts
```diff
@@ -34,36 +33,64 @@ export class GearService {
...
     this.balanceToTransfer = new BN(config.gear.balanceToTransfer);
-    if (await this.accountBalanceIsSmall()) {
-      this.transferBalance(this.account.address, this.rootAccount, this.accountBalance);
+    if (await this.isSmallAccountBalance()) {
+      await this.transferBalance(this.account.address, this.rootAccount, this.accountBalance);
+    }
+  }
+
+  public async transferBalance(to: string, from: KeyringPair, balance = this.accountBalance) {
+    await this.api.balance.transfer(to, balance);
```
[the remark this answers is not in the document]
Almazatun: fixed

PR #759, opened 2022-06-22
Refactor program page

Pinkdesu:
- Refactor program page


website/frontend/src/components/pages/Program/children/ProgramInfo/MetaData/MetaData.tsx
```diff
@@ -0,0 +1,45 @@
...
+        }
+      }
+    }
+
+    return items;
+  }, [metadata]);
+
+  return (
+    <div className={styles.fields}>
+      {metaFields.map((item) => (
```
aderrod94, line 38: I suggest to create a function which return result of the map() and call it here. As for me, **jsx** will be more clearly

osipov-mit (approved):

PR #763, opened 2022-06-24
Update api examples

osipov-mit:


nikitayutanov (approved):

api/examples/src/message/claimValue.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const alice = await GearKeyring.fromSuri('//Alice');
+
+  const mailbox = await api.mailbox.read(GearKeyring.decodeAddress(alice.address));
+
+  const messageWithValue = mailbox.find(([_, message]) => parseInt(message.value) > 0);
+
+  console.log(messageWithValue);
+
+  api.mailbox.claimValue.submit(messageWithValue[0][1]);
```
Almazatun, line 14: It might be better to create a variable for this `messageWithValue[0][1]`. It will be a little bit easier to understand what it is exactly.

```js
const someVariable = messageWithValue[0][1];
```

api/examples/src/message/claimValue.ts
```diff
@@ -0,0 +1,39 @@
...
+  try {
+    await new Promise((resolve, reject) => {
+      api.mailbox.claimValue.signAndSend(alice, ({ events, status }) => {
+        console.log(`STATUS: ${status.toString()}`);
+        if (status.isFinalized) resolve(status.asFinalized);
+        events.forEach(({ event }) => {
+          if (event.method === 'UserMessageRead') {
+            console.log(event.toHuman());
+          } else if (event.method === 'ExtrinsicFailed') {
```
Almazatun, line 23: What do think about this way?

Instead of using `else if` just add another `if`

```js
 if (event.method === 'UserMessageRead') {
        console.log(event.toHuman());
  }
  
  if (event.method === 'ExtrinsicFailed') {
      reject(api.getExtrinsicFailedError(event).docs.join('/n'))
  }

File: api/examples/src/message/waitForReply.ts [hunk 3]
```diff
@@ -0,0 +1,22 @@
...
+export function waitForReply(api: GearApi, programId: string): (messageId: string) => Promise<UserMessageSent> {
+  const messages = {};
+  api.query.system.events((events) => {
+    events.forEach(({ event }) => {
+      if (api.events.gear.UserMessageSent.is(event)) {
+        const {
+          data: {
+            message: { source, reply },
+          },
+        } = event as UserMessageSent;
```
Almazatun, line 12: It is possible to set the type before without using `as`?

```js
const { some key  }: UserMessageSent = event;
```

osipov-mit: It's impossible

Almazatun (approved):

api/examples/src/program/waitForInit.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function waitForInit(api: GearApi, programId: string): Promise<UnsubscribePromise> {
+  let messageId: Hex;
+  return new Promise((resolve, reject) => {
+    const unsub = api.query.system.events((events) => {
+      events.forEach(({ event }) => {
+        switch (event.method) {
+          case 'MessageEnqueued':
+            const meEvent = event as MessageEnqueued;
+            if (meEvent.data.destination.eq(programId) && meEvent.data.entry.isInit) {
```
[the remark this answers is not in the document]
osipov-mit: I don't want it because of naming. it will be like `isDestinationEqualsIdAndEntryIsInit`.

PR #785, opened 2022-07-01
Update and refactor kafka in @gear-js/event-listener

Almazatun:
Resolves # .

@reviewer-or-team

### DIFF
--- a/website/events-listener/src/kafka/producer.ts
+++ b/website/events-listener/src/kafka/producer.ts
@@ -1,71 +1,62 @@
-import { kafkaLogger } from '@gear-js/common';
-
-import { Kafka, Producer } from 'kafkajs';
-import config from '../config/configuration';
+import { Admin, Producer } from 'kafkajs';
 import { eventListenerLogger } from '../common/event-listener.logger';
+import { initKafka } from './init-kafka';
 
-export class KafkaProducer {
-  readonly kafka: Kafka;
-  readonly producer: Producer;
+const producer: Producer = initKafka.producer();
 
-  constructor() {
-    this.kafka = new Kafka({
-      clientId: config.kafka.clientId,
-      brokers: config.kafka.brokers,
-      logCreator: kafkaLogger,
-      sasl: {
-        mechanism: 'plain',
-        username: config.kafka.sasl.username,
-        password: config.kafka.sasl.password,
-      },
-    });
-    this.producer = this.kafka.producer();
-  }
+async function connect(): Promise<void> {
+  await producer.connect();
+  eventListenerLogger.info('Producer is connected');
+}
 
-  async createTopic(topic: string) {
-    const admin = this.kafka.admin();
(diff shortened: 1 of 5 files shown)


website/events-listener/src/kafka/producer.ts
```diff
@@ -1,71 +1,62 @@
...
+        waitForLeaders: true,
+        topics: [{ topic }],
+      });
+      eventListenerLogger.info(`Kafka producer: Topic ${topic} created`);
+    } else {
+      eventListenerLogger.warn(`Kafka producer: Topic ${topic} already existed`);
     }
+  } catch (error) {
+    eventListenerLogger.error(`Kafka producer: ${error}`);
     await admin.disconnect();
```
osipov-mit, line 52: `admin.disconnected()` should be called in any case, not only in case of an error

Almazatun: @osipov-mit  Do you mean like this way ?

```javascript
async function createKafkaTopic(admin: Admin, topic: string): Promise<void> {
  try {
    const kafkaTopics = await admin.listTopics();
    if (!isTopicAlreadyExist(kafkaTopics, topic)) {
      await admin.createTopics({
        waitForLeaders: true,
        topics: [{ topic }],
      });
      eventListenerLogger.info(`Kafka producer: Topic ${topic} created`);
    } else {
      eventListenerLogger.info(`Kafka producer: Topic ${topic} already existed`);
    }
  } catch (error) {
    eventListenerLogger.error(`Kafka producer: ${error}`);
    eventListenerLogger.info(`Admin is disconnected`);
    throw error;
  }
  
  await admin.disconnect();
}
```

osipov-mit: > @osipov-mit `amin.disconnected()` was there before in `catch` area. I didn't change it. In `master` branch currently the same in `catch` scope.

No, it was in catch and in after try/catch area. But I think there is no need to keep it in catch area. Because in that case application will go down and admin will be disconnected anyway

osipov-mit: I mean
```javascript
async function createKafkaTopic(admin: Admin, topic: string): Promise<void> {
  try {
    const kafkaTopics = await admin.listTopics();
    if (!isTopicAlreadyExist(kafkaTopics, topic)) {
      await admin.createTopics({
        waitForLeaders: true,
        topics: [{ topic }],
      });
      eventListenerLogger.info(`Kafka producer: Topic ${topic} created`);
    } else {
      eventListenerLogger.info(`Kafka producer: Topic ${topic} already existed`);
    }
  } catch (error) {
    eventListenerLogger.error(`Kafka producer: ${error}`);
    throw error;
  }
  
  await admin.disconnect();
  eventListenerLogger.info(`Admin is disconnected`);
}
```

osipov-mit (approved):

PR #864, opened 2022-07-21
Refactor `@gear-js/api`

osipov-mit:


api/src/utils/signature.ts
```diff
@@ -0,0 +1,9 @@
+import { signatureVerify } from '@polkadot/util-crypto';
+
+export function signatureIsValid(publicKey: string, signature: string, message: string) {
+  if (signatureVerify(message, signature, publicKey).isValid) {
```
Pinkdesu, line 4: return signatureVerify(message, signature, publicKey).isValid

api/src/utils/signature.ts
```diff
@@ -0,0 +1,5 @@
+import { signatureVerify } from '@polkadot/util-crypto';
+
+export function signatureIsValid(publicKey: string, signature: string, message: string) {
```
nikitayutanov, line 3:
```suggestion
export function isSignatureValid(publicKey: string, signature: string, message: string) {
```

nikitayutanov (approved):

PR #905, opened 2022-07-30
Get test balance from multiply `@gear-js/test-balance` instances

Almazatun:
Resolves # . 

@reviewer-or-team


idea/api-gateway/src/common/genesis-hash-map.ts
```diff
@@ -0,0 +1,3 @@
+const genesisHashMap = new Map<string, string>();
```
osipov-mit, line 1: 1. What is the point to keep this and scheduler in common package if it used only in api-gateway?
2. Maybe it's better to use Set instead Map?

idea/api-gateway/src/common/scheduler-genesis-hashes.ts
```diff
@@ -0,0 +1,32 @@
...
+      //   const SEVENTY_SECONDS = 70 * 1000;
+      //   setTimeout(async () => {
+      //     await kafkaProducer.sendByTopic(KAFKA_TOPICS.TEST_BALANCE_SERVICES, 'testBalanceServices');
+      //   }, SEVENTY_SECONDS);
+      // }
+    }
+  };
+}
+
+function getCronRunTime(): string {
```
osipov-mit, line 27: Why is this a function? I think it might be constant

idea/api-gateway/src/kafka/consumer.ts
```diff
@@ -12,10 +14,17 @@ async function connect(): Promise<void> {
...
-      if (resultFromService) await resultFromService(JSON.parse(message.value.toString()));
-      deleteKafkaEvent(correlationId);
+      if (isIncludeCorrelationId(message)) {
+        const correlationId = message.headers.kafka_correlationId.toString();
+        const resultFromService = kafkaEventMap.get(correlationId);
+        if (resultFromService) await resultFromService(JSON.parse(message.value.toString()));
+        deleteKafkaEvent(correlationId);
+      }
+
+      if (!isIncludeCorrelationId(message)) {
```
osipov-mit, line 24: What is the point to check it 2 times? Use just else

idea/api-gateway/src/common/genesis-hash-map.ts
Almazatun: @osipov-mit What do mean by "if it used only in `api-gateway` ?

Currently `genesisHashMap` declared in `api-gateway`. Each of the services also can have a `common` folder for sharing data inside service no between services.

I think so with you. It will be better to use `Set` instead of `Map`.

osipov-mit: Ah, yeah, sorry. I misunderstood

idea/api-gateway/src/json-rpc/json-rpc-request.handler.ts
```diff
@@ -24,6 +24,14 @@ async function executeProcedure(procedure: IRpcRequest): Promise<IRpcResponse> {
     apiGatewayLogger.error(`${API_GATEWAY}:${JSON.stringify(JSONRPC_ERRORS.MethodNotFound)}`);
     return getResponse(procedure, JSONRPC_ERRORS.MethodNotFound.name);
   }
+
+  if (procedure.method === KAFKA_TOPICS.TEST_BALANCE_GET
+      && genesisHashesCollection.has(procedure.params.genesis)) {
+    const { params: { genesis } } = procedure;
+    const genesisFromCollection = Array.from(genesisHashesCollection.values()).find(gen => gen === genesis);
```
osipov-mit, line 31: What is it for?

idea/api-gateway/src/json-rpc/json-rpc-request.handler.ts
```diff
@@ -24,6 +24,14 @@ async function executeProcedure(procedure: IRpcRequest): Promise<IRpcResponse> {
     apiGatewayLogger.error(`${API_GATEWAY}:${JSON.stringify(JSONRPC_ERRORS.MethodNotFound)}`);
     return getResponse(procedure, JSONRPC_ERRORS.MethodNotFound.name);
   }
+
+  if (procedure.method === KAFKA_TOPICS.TEST_BALANCE_GET
```
osipov-mit, line 28: What happens if the 1st condition is true and the 2nd is false?

idea/api-gateway/src/kafka/consumer.ts
```diff
@@ -12,10 +14,15 @@ async function connect(): Promise<void> {
 async function run(): Promise<void> {
   await consumer.run({
     eachMessage: async ({ message }) => {
-      const correlationId = message?.headers?.kafka_correlationId.toString();
-      const resultFromService = kafkaEventMap.get(correlationId);
-      if (resultFromService) await resultFromService(JSON.parse(message.value.toString()));
-      deleteKafkaEvent(correlationId);
+      if (isIncludeCorrelationId(message)) {
```
osipov-mit, line 17: I think it's better to check topic name instead correlation id

idea/api-gateway/src/kafka/producer.ts
```diff
@@ -10,19 +11,22 @@ async function connect(): Promise<void> {
...
-      },
-    ],
+    messages: [getSendKafkaMessage(topic, params, correlationId)],
   });
 }
 
+function getSendKafkaMessage(topic: string, params: KafkaParams | string, correlationId?: string): Message {
+  const result: Message = { value: transformToSting(params), headers: {} };
+
+  if (correlationId) {
```
osipov-mit, line 24: I don't see reason to send message without correlation id

idea/api-gateway/src/kafka/producer.ts
```diff
@@ -10,19 +11,22 @@ async function connect(): Promise<void> {
...
-          kafka_correlationId: correlationId,
-          kafka_replyTopic: `${topic}.reply`,
-        },
-      },
-    ],
+    messages: [getSendKafkaMessage(topic, params, correlationId)],
   });
 }
 
+function getSendKafkaMessage(topic: string, params: KafkaParams | string, correlationId?: string): Message {
```
osipov-mit, line 21: Weird name

idea/api-gateway/src/services/api-gateway.service.ts
```diff
@@ -15,7 +15,6 @@ export const apiGatewayService = {
         return getResponse(body, JSONRPC_ERRORS.Forbidden.name);
       }
     }
-    const response = await jsonRpcRequestHandler(body);
-    return response;
+    return await jsonRpcRequestHandler(body);
```
osipov-mit, line 18: Actually we can return result without `await`

osipov-mit (changes requested):

osipov-mit:
Pls, try to add tests for the new functionality

idea/api-gateway/src/kafka/producer.ts
Almazatun: @osipov-mit  `getMessageBody` what do you think ?

osipov-mit: Maybe create instead get?

idea/api-gateway/src/kafka/producer.ts
osipov-mit: Why is this conversation marked as resolved?

idea/api-gateway/src/config/configuration.ts
```diff
@@ -22,4 +22,8 @@ export default () => ({
       password: checkEnv('KAFKA_SASL_PASSWORD'),
     },
   },
+  // every 1 hour "0 * * * *"
+  cron: {
+    time: checkEnv('CRON_TIME') || '0 * * * *',
```
osipov-mit, line 27: In this case the default value will not work, because `checkEnv` will throw an error if variable isn't unspecified

idea/api-gateway/src/common/enums/genesis-hash.enum.ts
```diff
@@ -0,0 +1,5 @@
+enum GenesisRpc {
+    valid = 'genesis.valid'
```
osipov-mit, line 2: `genesis.valid` is strange choice. It doesn't reflect the necessary meaning. Maybe `testBalance.available` will be better

osipov-mit: And why you use separate enum for this method? I would store it together with the rest of the rpc methods

idea/tests/src/utils.ts
```diff
@@ -1,7 +1,7 @@
-export function sleep() {
+export function sleep(time?: number) {
```
osipov-mit, line 1:
```suggestion
export function sleep(time: number = 2000) {
```

idea/tests/src/utils.ts
```diff
@@ -1,7 +1,7 @@
...
   return new Promise((resolve) => {
     setTimeout(() => {
       resolve(0);
-    }, 2000);
+    }, time ? time : 2000);
```
osipov-mit, line 5:
```suggestion
    }, time);
```

idea/tests/src/e2e/e2e.test.ts
```diff
@@ -97,4 +97,8 @@ describe('testBalance', () => {
   test('testBalance.get request', async () => {
     expect(await getTestBalance(genesis)).toBeTruthy();
   });
+  test('genesis.valid request', async () => {
+    await sleep(50_000);
```
osipov-mit, line 101: What's the point of sleeping here?

idea/test-balance/src/services/genesis-hash/genesis-hash.service.ts
```diff
@@ -0,0 +1,41 @@
...
+        result = { error: JSONRPC_ERRORS.TransferLimitReached.name };
+      }
+    } catch (error) {
+      logger.error(error.message, error.stack);
+      result = { error: JSONRPC_ERRORS.InternalError.name };
+    }
+
+    await sendReply(message, JSON.stringify(result));
+  },
+  async sendApiGateway(): Promise<void> {
```
osipov-mit, line 29: `sendApiGateway`? What does it mean?

idea/test-balance/src/kafka/producer.ts
```diff
@@ -8,16 +8,21 @@ async function connect(): Promise<void> {
...
+    messages: [createMessageBody(value, correlationId)],
   });
 }
 
+function createMessageBody(value: string, correlationId?: string): Message {
+  const result: Message = { value, headers: {} };
+
+  if (correlationId) {
+    result.headers = { kafka_correlationId: correlationId };
+    return result;
```
osipov-mit, line 23: Why return the result here?

osipov-mit (changes requested):

idea/api-gateway/src/common/enums/genesis-hash.enum.ts
osipov-mit: We can create 2 enums, one with all Kafka methods and another one with all api methods.

idea/test-balance/src/services/genesis-hash/genesis-hash.service.ts
osipov-mit: Yeah, I see the main idea, but sendApiGateway sounds strange as if you are sending some apiGateway. Maybe `sendGenesis` would be better or smth else

idea/api-gateway/src/common/enums/genesis-hash.enum.ts
Almazatun: In `@gear-js/common` we already have `API_METHODS` `enum`. Maybe the new `method` add to this `enum`. What do you think @osipov-mit 

https://github.com/gear-tech/gear-js/blob/master/idea/common/src/enums/api-methods.enum.ts#L1

osipov-mit: In my opinion its better to use separate enums for kafka topics and api methods (and describe all the methods there).

osipov-mit (approved):

PR #917, opened 2022-08-09
Fix saving message payload

osipov-mit:


idea/data-storage/src/message/message.service.ts
```diff
@@ -81,8 +80,8 @@ export class MessageService {
     }
   }
 
-  public async updateMessagesData(updateMessagesParams: UpdateMessagesParams): Promise<void> {
-    const promises = updateMessagesParams.params.map((updateMessageData) => {
+  public async updateMessagesData(updateMessagesParams: UpdateMessageData[]): Promise<void> {
```
Almazatun, line 83:
```suggestion
  public async updateMessagesData(updateMessagesParams: UpdateMessageData[]): Promise<void> {
```

osipov-mit: Hm?

Almazatun:
@osipov-mit Also looks like `UpdateMessagesParams` is unused anymore from `@gear-js/common`. Does that mean we can delete it?

`UpdateMessagesParams`
https://github.com/gear-tech/gear-js/blob/do-fix-message-update/idea/common/src/interfaces/api-requests.ts#L38

`ExtrinsicsResult`
https://github.com/gear-tech/gear-js/blob/do-fix-message-update/idea/events-listener/src/gear/types/index.ts#L11

Almazatun (approved):

PR #949, opened 2022-08-19
Add supply chain app

nikitayutanov:


apps/supply-chain/src/pages/home/create/address-form/AddressForm.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { isValidHex } from 'utils';
+import commonStyles from '../Create.module.scss';
+import styles from './AddressForm.module.scss';
+
+type Props = {
+  id: string;
+  onSubmit: (value: { nftProgramId: Hex; ftProgramId: Hex }) => void;
+};
+
+const initialValues = {
```
Pinkdesu, line 15: Move it in consts, helpers

apps/supply-chain/src/pages/home/create/address-form/AddressForm.tsx
```diff
@@ -0,0 +1,51 @@
...
+function AddressForm({ id, onSubmit }: Props) {
+  const { api } = useApi();
+
+  const form = useForm({ initialValues, validate });
+  const { getInputProps, setFieldError } = form;
+
+  const nftInputClassName = clsx(commonStyles.input, styles.nftInput);
+
+  const handleSubmit = form.onSubmit(async ({ nftProgramId, ftProgramId }) => {
+    const isNftExists = await api.program.exists(nftProgramId);
```
Pinkdesu, line 30: use Promise.all

apps/supply-chain/src/pages/home/program/Program.tsx
```diff
@@ -0,0 +1,95 @@
...
+                        action={getAction(action)}
+                        label={getLabel(action)}
+                        name={getName(action)}
+                        onSubmit={handleSubmit}
+                      />
+                    )}
+                  </>
+                ) : (
+                  <SelectText value={selectTextValue} />
+                )}
```
Pinkdesu, line 83: Maybe move part of the logic into a separate component, difficult to read

apps/supply-chain/src/pages/home/create/address-form/AddressForm.tsx
nikitayutanov: I prefer to not separate small encapsulated values

nikitayutanov: Placeholder values are gone. 
As for utils, I prefer to not separate small encapsulated values

apps/supply-chain/src/pages/home/program/Program.tsx
nikitayutanov: Suggest what component and what logic should be inside.
I agree that the way it is rn is a bit complicated, but don't see better solution

apps/supply-chain/src/pages/home/create/address-form/AddressForm.tsx
nikitayutanov: There's a need to get different errors if either `nft` or `ft` program doesn't exist.

nikitayutanov: Idk, current solution looks clearer to me:
image

Pinkdesu (approved):

PR #958, opened 2022-08-24
Add useUploadProgram hook

nikitayutanov:


aderrod94 (approved):

utils/gear-hooks/src/hooks/api/useCalculateGas/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import { AccountContext, ApiContext } from 'context';
+import { CalculateGas, HandleOptions, InitOptions } from './types';
+
+function useCalculateGas(
+  programBuffer: Buffer | undefined,
+  metadata?: Metadata | undefined,
+  options?: InitOptions,
+): CalculateGas;
+function useCalculateGas(
+  programIdOrBuffer: Hex | Buffer | undefined,
```
osipov-mit, line 13: Did you mean codeId? If it is used for `createProgram` extrinsic then this should be called codeId, not programId

utils/gear-hooks/src/hooks/api/useUploadProgram/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+
+            if (pcEvent.data.id.eq(programId) && pcEvent.data.change.isActive) resolve(ProgramStatus.Success);
+
+            break;
+          }
+          default:
+            break;
+        }
+      });
+    });
```
osipov-mit, line 54: I found a way to make it a little easier

```javascript
function checkInit(api) {
  let processedMessages = new Map();
  const unsubPromise = api.query.system.events((events) => {
    events.forEach(({ event: { method, data } }) => {
      if (method === 'MessagesDispatched') {
        for (const [id, status] of data.statuses) {
          processedMessages.set(id.toHex(), status.isSuccess);
        }
      }
    });
  });

  return (messageId) => {
    unsubPromise.then((unsub) => unsub());
    if (!processedMessages.has(messageId)) {
      throw Error('Message was not processed')
    }
    const status = processedMessage.get(messageId);
    return status ? resolve(ProgramStatus.Success) : resolve(ProgramStatus.Failed);
  }
};
```

osipov-mit: In this case you need to pass `messageId`, from `MessageEnqueued` event instead of `programId`

nikitayutanov: I'd need to get `messageId` from `signAndSend`'s callback and not from it's returned promise, therefore it's not really easier.

osipov-mit (approved):

PR #960, opened 2022-08-29
Add nft marketplace indexer

osipov-mit:


apps/indexers/nft-marketplace/src/handlers/marketplace.ts
```diff
@@ -0,0 +1,168 @@
...
+        to: owner,
+        timestamp: BigInt(timestamp),
+        price: BigInt(price),
+      });
+      token.owner = await store.findOneBy(Account, { id: newOwner });
+      // token.transfers.push(transfer);
+      token.price = null;
+      await store.save(token);
+      await store.save(auction);
+      await store.save(transfer);
```
Almazatun, line 123: it is possible to resolve asynchronous processes with `Promise.all` here

And some others below?

apps/indexers/nft-marketplace/src/handlers/nft.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+  const variant = Object.keys(payload)[0];
+
+  const nftContract = await store.findOneBy(NftContract, { id: message.source });
+  const dbTokenId = `${message.source}-${payload[variant].tokenId}`;
+
+  const variants: Record<string, Function> = {
+    minted: async ({ tokenMetadata: { name, reference, media, description }, owner, tokenId }: IToken) => {
+      const account = await getAcc(owner, store);
+      await store.save(
```
Almazatun, line 31: @osipov-mit  It's up to you. But in my opinion, looks better. 

```js
const token = new Token()
await  store.save(token)
```

apps/indexers/nft-marketplace/src/indexNft.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  let {
+    data: { batch },
+  } = await archiveReq(config.squid.archiveUrl, query);
+
+  console.log(JSON.stringify(batch, undefined, 4));
+
+  if (batch.length === 0) return;
+
+  for (const { header, events } of batch) {
```
Almazatun, line 28: @osipov-mit  What do think?

```js
const promises = [];

for (const { header, events } of batch) {
  for (const { args, name, id } of events) {
      if (name !== 'Gear.UserMessageSent') continue;
         promises.push(handleNftEvent(args.message, store, id, Date.parse(header.timestamp)));
       }
  }
  
  await Promise.all(promises)
  ```

File: apps/indexers/nft-marketplace/src/model/generated/account.model.ts [hunk 4]
```diff
@@ -0,0 +1,15 @@
+import {Entity as Entity_, Column as Column_, PrimaryColumn as PrimaryColumn_, OneToMany as OneToMany_} from "typeorm"
+import {Token} from "./token.model"
+
+@Entity_()
```
Almazatun, line 4: What does the reason to use `Entity_()` instead of `Entity()` ?

apps/indexers/nft-marketplace/src/model/generated/marshal.ts
```diff
@@ -0,0 +1,134 @@
...
+    },
+}
+
+
+// credit - https://github.com/Urigo/graphql-scalars/blob/91b4ea8df891be8af7904cf84751930cc0c6613d/src/scalars/iso-date/validator.ts#L122
+const RFC_3339_REGEX =
+    /^(\d{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60))(\.\d{1,})?([Z])$/
+
+
+function isIsoDateTimeString(s: string): boolean {
```
Almazatun, line 73:
```suggestion
function isISODateTimeString(s: string): boolean {
```

apps/indexers/nft-marketplace/src/model/generated/marshal.ts
```diff
@@ -0,0 +1,134 @@
...
+        if (Buffer.isBuffer(value)) {
+            return '0x' + value.toString('hex')
+        } else {
+            return '0x' + Buffer.from(value.buffer, value.byteOffset, value.byteLength).toString('hex')
+        }
+    },
+}
+
+
+export function fromList<T>(list: unknown, f: (val: unknown) => T): T[] {
```
Almazatun, line 107:
```suggestion
export function fromList<T>(list: unknown, callback: (val: unknown) => T): T[] {
```

apps/indexers/nft-marketplace/src/handlers/marketplace.ts
eldargab: Database queries are executed one-by-one in a single transaction. So, `Promise.all` would just create additional promises slightly degrading performance.

apps/indexers/nft-marketplace/src/model/generated/account.model.ts
eldargab: This is an auto generated code. `_` is a hack to avoid name collisions.

apps/indexers/nft-marketplace/schema.graphql
```diff
@@ -0,0 +1,58 @@
...
+}
+
+type Bid @entity {
+  auction: Auction!
+  price: BigInt!
+}
+
+type Auction @entity @index(fields: ["nftContract", "token"]){
+  token: Token! @index
+  nftContract: NftContract! @index
```
eldargab, line 18: Index on `nftContract` alone is redundant, because of `@index(fields: ["nftContract", "token"])`

apps/indexers/nft-marketplace/schema.graphql
```diff
@@ -0,0 +1,58 @@
...
+  auction: Auction
+  isListed: Boolean!
+  price: BigInt
+  transfers: [Transfer!]! @derivedFrom(field: "token")
+  offers: [Offer!]! @derivedFrom(field: "token")
+  burnt: Boolean!
+  
+}
+
+type Transfer @entity @index(fields: ["token"]){
```
eldargab, line 44: `@index(fields: ["token"])` is redundant.

apps/indexers/nft-marketplace/src/logger.ts
```diff
@@ -0,0 +1,16 @@
+import { createLogger, format, transports } from 'winston';
+import chalk from 'chalk';
+
+const { combine, printf } = format;
+
+const actionFormat = printf(({ message, label }) => `${chalk.magenta('[')}${label}${chalk.magenta(']')} ${message}`);
+
+const logger = createLogger({
```
eldargab, line 8: Probably you know, but there is a [built-in structured logger](https://github.com/subsquid/squid/tree/master/util/logger) available in mappings via `ctx.log`.

apps/indexers/nft-marketplace/src/model/generated/marshal.ts
osipov-mit: autogenerated file

apps/indexers/nft-marketplace/src/model/generated/marshal.ts
osipov-mit: auto generated file

apps/indexers/nft-marketplace/src/logger.ts
osipov-mit: Yeah, this part has been removed

apps/indexers/nft-marketplace/src/indexNft.ts
osipov-mit: Refactored

apps/indexers/nft-marketplace/src/archiveReq.ts
```diff
@@ -0,0 +1,20 @@
+import fetch from 'node-fetch';
+import { IBatchResponse } from './types/batch';
+
+export async function archiveReq(url: string, query: string): Promise<IBatchResponse> {
```
dzhelezov, line 4: We have a small util lib for gql requests: `https://github.com/subsquid/squid/blob/master/util/util-internal-gql-request/src/gql-request.ts`

It has built-in retry with backoffs support

dzhelezov (approved):

Almazatun (approved):

eldargab (approved):
If performance still matters and there are a lot of `store.find()` calls issued per batch, I can suggest the following changes:

1. Modify `batchState` methods like `.burnToken()` or `.transferToken()` in such a way, that they would become synchronous and wouldn't do any work except queueing received data for further processing
2. Processing should be 2 phase. 
3. At first phase all missing database data should be fetched using as few queries as possible.
4. At second phase we do actual processing as it is implemented now, except it becomes completely synchronous

osipov-mit:
> If performance still matters and there are a lot of `store.find()` calls issued per batch, I can suggest the following changes:
> 
> 1. Modify `batchState` methods like `.burnToken()` or `.transferToken()` in such a way, that they would become synchronous and wouldn't do any work except queueing received data for further processing
> 2. Processing should be 2 phase.
> 3. At first phase all missing database data should be fetched using as few queries as possible.
> 4. At second phase we do actual processing as it is implemented now, except it becomes completely synchronous

Okay, thank you! I'll think about it

PR #1096, opened 2022-11-17
Fix obsolete local programs display

nikitayutanov:


ereburg (approved):

idea/frontend/src/hooks/useLocalProgramsFilter.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const filterLocalPrograms = () => {
+    const genesis = api.genesisHash.toHex();
+
+    api.program.allUploadedPrograms().then((chainProgramIDs) =>
+      PROGRAMS_LOCAL_FORAGE.iterate((program: IProgram) => {
+        const isProgramInChain = chainProgramIDs.includes(program.id);
+        const isProgramFromChain = genesis === program.genesis;
+
+        if (!isProgramInChain || !isProgramFromChain) PROGRAMS_LOCAL_FORAGE.removeItem(program.id);
```
osipov-mit, line 21:
```suggestion
        if (!isProgramInChain && !isProgramFromChain) PROGRAMS_LOCAL_FORAGE.removeItem(program.id);
```
I think you only need to remove the programs related to connected chain. What do you think?

nikitayutanov: In this case if two programs have the same id, but different genesis, it won't be deleted.
There's no additional genesis check on programs render.

osipov-mit (approved):

PR #1183, opened 2023-02-03
Add Tamagotchi dapp

ereburg:


osipov-mit:
Must be moved to repo with tamagotchi contract. @ereburg @nikitayutanov

PR #1186, opened 2023-02-07
Fix code `meta`

Almazatun:
Resolves # .

- Update `meta` type in  `@gear-js/common`.
-  Refactor and update `GearEventListener` service.

@reviewer-or-team

### DIFF
--- a/idea/data-storage/src/common/helpers/get-meta-hash.ts
+++ b/idea/data-storage/src/common/helpers/get-meta-hash.ts
@@ -1,8 +1,8 @@
 import { HexString } from '@polkadot/util/types';
-import { GearProgram } from '@gear-js/api';
-import { GearCode } from '@gear-js/api/Code';
 
-export async function getMetaHash<T extends GearProgram | GearCode>(tx: T, hex: HexString): Promise<null | HexString> {
+import { GearApiModule } from '../types';
+
+export async function getMetaHash(tx: GearApiModule, hex: HexString): Promise<null | HexString> {
   try {
     const metaHash = await tx.metaHash(hex);
     if(metaHash) return metaHash;
--- a/idea/data-storage/src/common/helpers/get-meta-hash.ts
+++ b/idea/data-storage/src/common/helpers/get-meta-hash.ts
@@ -1,0 +1,13 @@
+import { HexString } from '@polkadot/util/types';
+import { GearProgram } from '@gear-js/api';
+import { GearCode } from '@gear-js/api/Code';
+
+export async function getMetaHash<T extends GearProgram | GearCode>(tx: T, hex: HexString): Promise<null | HexString> {
+  try {
+    const metaHash = await tx.metaHash(hex);
+    if(metaHash) return metaHash;
+  } catch (error) {
+    console.log(`${new Date}, getMetaHash: ${error}`);
+    return null;
+  }
+}
--- a/idea/data-storage/src/gear/gear-event-listener.ts
+++ b/idea/data-storage/src/gear/gear-event-listener.ts
@@ -11,7 +11,7 @@ import { plainToClass } from 'class-transformer';
 import { ProgramService } from '../program/program.service';
 import { MessageService } from '../message/message.service';
 import { CodeService } from '../code/code.service';
-import { getPayloadAndValue, getPayloadByGearEvent } from '../common/helpers';
+import { getMetaHash, getPayloadAndValue, getPayloadByGearEvent } from '../common/helpers';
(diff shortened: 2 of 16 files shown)


idea/data-storage/src/common/helpers/get-meta-hash.ts
```diff
@@ -0,0 +1,13 @@
+import { HexString } from '@polkadot/util/types';
+import { GearProgram } from '@gear-js/api';
+import { GearCode } from '@gear-js/api/Code';
+
+export async function getMetaHash<T extends GearProgram | GearCode>(tx: T, hex: HexString): Promise<null | HexString> {
```
osipov-mit, line 5: What's the point to use generic here?

idea/data-storage/src/gear/gear-event-listener.ts
```diff
@@ -276,30 +257,57 @@ export class GearEventListener {
...
             expiration: change.isActive ? change.asActive.expiration.toString() : null,
             uploadedBy: tx.signer.inner.toHex(),
-          });
+            meta: null,
+          };
+
+          const metaHash = await getMetaHash(this.api.code, id.toHex());
+
+          if(metaHash) {
+            const meta = await this.metaService.getByHash(metaHash);
```
osipov-mit, line 274: Perhaps it would better to make method `getaOrCreate` that returns a Meta object anyway?

osipov-mit (changes requested):

idea/data-storage/src/common/helpers/get-meta-hash.ts
Almazatun: Replaced with created type `GearApiModule` instead of this `genric` type here.

idea/data-storage/src/meta/meta.service.ts
```diff
@@ -26,10 +26,28 @@ export class MetaService {
     private gearEventListener: GearEventListener,
   ) {}
 
+  public async getByHashApi(hash: string): Promise<Meta> {
```
osipov-mit, line 29: I'm sure that `getOrCreate` is musch better

idea/data-storage/src/gear/gear-event-listener.ts
```diff
@@ -276,30 +257,45 @@ export class GearEventListener {
...
+
+          if(metaHash) {
+            updateCodeInput.meta = await this.metaService.getByHashApi(metaHash);
+          }
+
+          codes.push(updateCodeInput);
         } else {
           const codeId = generateCodeHash(tx.args[0].toHex());
           const code = await this.codeRepository.get(codeId, this.genesis);
+          const metaHash = await getMetaHash(this.api.code, codeId);
```
osipov-mit, line 281: If you need to get meta under both conditions, I assume that it's better to do it once before if/else

osipov-mit (changes requested):

idea/data-storage/src/meta/meta.service.ts
Almazatun: I don't think so. It's bad practice to use `or` when naming `function or method` in my opinion. But also not a good idea when function does more than one thing. For that reason before I'm not combining two actions in one method.

For that situation much better to use `get` but inside `method` we know that used two separate methods.

osipov-mit: Anyway getByHashApi means nothing

Almazatun: Why nothing? I didn't get you. Do you think it's a good practice when `function` do more than one thing?

osipov-mit: I think that name of function should be meaningful and reflect what function does. What does HashApi mean?

Almazatun: The method not called `HashApi` is correct name is `getByHashApi`. The `Api` prefix I have added for recognizing that method is used in `gearEventListerner`.

If some function will do more than one thing it will look like `getOrCreateOrUpdate`. May be meaningful but is not a good naming of a function and not a good idea to create like this function.

osipov-mit: In Rust, for example, you can defentely find many standart functions called `get_or_create` or `get_or_insert` etc. So I'm not agree with you that it's bad way. Of course more often it's not good to create functions that do a lot of actions but in this particulare case it looks justified.

nikitayutanov (approved):

osipov-mit (approved):

PR #1221, opened 2023-03-07
Consecutive transfers

osipov-mit:
Refactor test-balance service.
- Process requests to get test balance in sequence to avoid several transactions with the same account nonce
- Fix reconnection to the node


Almazatun (approved):

Almazatun:
Pls, add a description of what you did in this PR @osipov-mit

idea/test-balance/src/gear/index.ts
```diff
@@ -1,85 +1,57 @@
...
   to: string,
-  from: KeyringPair = accountGR,
-  balance: BN = balanceToTransferGR,
+  from: KeyringPair = tbAccount,
+  balance: BN = balanceToTransfer,
 ): Promise<ResponseTransferBalance> {
+  logger.info(`Transfer value ${balance.toNumber()} from ${from.address} to ${to}`);
   try {
-    await transfer(from, to, balance);
+    await transfer(to, from, balance);
```
Almazatun, line 39: @osipov-mit What do you think before calling function `transfer` inside `transferBalance` to check 
with flag function `isSmallAccountBalance()` balance of the root account to avoid error that we get on our `idea`.

error - `[TEST_BALANCE_GEAR] [{"module":{"index":5,"error":"0x02000000"}},{"weight":{"refTime":266899000,"proofSize":0},"class":"Normal","paysFee":"Yes"}]`

Also, we can create `JSON_RPC` error for that case.

idea/tests/src/e2e/testBalance.ts
```diff
@@ -13,6 +13,33 @@ export async function getTestBalance(genesis: string): Promise<Passed> {
   return true;
 }
 
+export async function getTestBalanceSeveralTimesAtATime(genesis: string) {
+  const promises = Promise.all([
```
Almazatun, line 17: @osipov-mit What do you think?

```javascript

  const [res_1, res_2, res_3] = await Promise.all([request('testBalance.get', {
      genesis,
      address: '5HpG9w8EBLe5XCrbczpwq5TSXvedjrBGCwqxK1iQ7qUsSWFc',
      token: 'test_token',
    }),
    request('testBalance.get', {
      genesis,
      address: '5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y',
      token: 'test_token',
    }),
    request('testBalance.get', {
      genesis,
      address: '5DAAnrj7VHTznn2AWBemMuyBwZWs6FNFjdyVXUeYum3PTXFy',
      token: 'test_token',
    }),
  ]);

  expect(res_1).to.have.own.property('result');
  expect(res_2).to.have.own.property('result');
  expect(res_3).to.have.own.property('result');
  return true;
```

idea/test-balance/src/common/transfer-balance-process.ts
```diff
@@ -4,21 +4,48 @@ import { transferService } from '../services/transfer.service';
 import { gearService } from '../gear';
 import { testBalanceLogger } from './test-balace.logger';
 import { producer } from '../rabbitmq/producer';
+import { EventEmitter } from 'node:events';
 
-export async function transferBalanceProcess(payload: any, correlationId: string): Promise<void> {
-  let result;
+export const requests: Array<{ payload: any; correlationId: string }> = [];
```
Almazatun, line 9: What do you think, maybe we can create a `type` by the name `RequestParams` in rabbitmq directory in the types folder, and use this type here. 

```javascript
interface RequestParams {
  payload: any;
  correlationId: string;
}
```

```javascript
const requests = RequestParams[] = [];
```

idea/test-balance/src/common/transfer-balance-process.ts
```diff
@@ -4,21 +4,48 @@ import { transferService } from '../services/transfer.service';
...
+    if (req) {
+      yield req;
+    }
+  }
+}
+
+export async function transferProcess(): Promise<void> {
+  for await (const { payload, correlationId } of transferGenerator()) {
+    let result;
+
```
Almazatun, line 36: Small suggestion

```javascript
const { address, genesis } = payload;

const isPossibleToTransfer = await transferService.isPossibleToTransfer(address, genesis);

const transferBalance = await gearService.transferBalance(address);

```

idea/test-balance/src/gear/index.ts
osipov-mit: Not a good idea actually. It will increase the time of request which has already increased in some cases due to sequential order of the requests. I have some thoughts on how to do it better but not here.

idea/test-balance/src/common/transfer-balance-process.ts
osipov-mit: doesn't matter

idea/test-balance/src/common/transfer-balance-process.ts
osipov-mit: yep

idea/test-balance/src/common/transfer-balance-process.ts
Almazatun: It matters for future people who will read your code easily not only for you

PR #1255, opened 2023-04-07
Fix getMeta and index messages

osipov-mit:
- Return meta as long as it has `hex` filed that isn't null
- Reindex block with a program not found


idea/indexer/src/common/form-response.middleware.ts
Almazatun, line 1: Why this file has a prefix with `middleware`? If it's not middleware.

idea/indexer/src/services/message.service.ts
```diff
@@ -71,7 +70,7 @@ export class MessageService {
     };
   }
 
-  public async create(messages: Message[]): Promise<IMessage[]> {
+  public async create(messages: Message[]): Promise<Message[]> {
     for (const m of messages) {
```
Almazatun, line 74: What is the advantage use a single letter? It makes code more readable and what kind of quality improves if using a single letter? In my opinion, it makes more decreases readable quality for other people who will look at this code.

idea/indexer/src/common/form-response.middleware.ts
osipov-mit: right

idea/indexer/src/services/message.service.ts
osipov-mit: It's good when code is readable and short as much as it possible. In this case it's absolutely clear what `m` means. I don't see any problem

Almazatun (approved):

PR #1260, opened 2023-04-11
Update indexer

osipov-mit:
- Index blocks in batches during ony-time-sync
- Add temporary state to decrease number of r/w opertaions to database

### DIFF
--- a/idea/indexer/src/common/helpers/get-payload-by-gear-event.ts
+++ b/idea/indexer/src/common/helpers/get-payload-by-gear-event.ts
@@ -57,26 +57,25 @@ function userMessageReadPayload({ id, reason }: UserMessageReadData): UserMessag
 }
 
 function programChangedPayload({ id, change }: ProgramChangedData): ProgramChangedInput {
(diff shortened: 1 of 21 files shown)


idea/indexer/src/common/helpers/get-payload-by-gear-event.ts
```diff
@@ -57,26 +57,25 @@ function userMessageReadPayload({ id, reason }: UserMessageReadData): UserMessag
...
-  }
-
-  if (change.isInactive) {
-    res.programStatus = ProgramStatus.EXITED;
-  }
-
-  if (change.isPaused) {
-    res.programStatus = ProgramStatus.EXITED;
-  }
+  const status = change.isActive
```
Almazatun, line 60: IMHO with `if` statement is easier to read instead of the steps `ternary operations`. Of course, it's up to you to refactor this part or not.

osipov-mit: These conditions are small, so I think it's okay in this case.

PR #1282, opened 2023-05-02
Add `meta-storage` service

Almazatun:
Resolves # .

- Add `@gear-js/meta-storage` service.
- Update `program` and `code` entities in `@gear-js/indexer`.
- Delete API methods `program.meta.add`, `program.meta.get`, `code.meta.get`, `code.meta.add`.
- Add API methods `meta.add` and `meta.get`
- Update `e2e` tests.
@reviewer-or-team


idea/common/src/interfaces/meta.ts
```diff
@@ -1,7 +1,7 @@
 import { ProgramMetadata } from '@gear-js/api';
 
 export interface IMeta {
-  id: number;
+  id: string;
```
osipov-mit, line 4: Let's name it `hash`

idea/meta-storage/src/services/meta/input/create-meta.input.ts
```diff
@@ -0,0 +1,4 @@
+export interface CreateMetaInput {
```
osipov-mit, line 1: I don't see the point of creating and using this interface

idea/meta-storage/src/services/meta/meta.service.ts
```diff
@@ -0,0 +1,51 @@
...
+import { MetaAlreadyExists, MetadataNotFound } from '../../common/errors';
+import { generateHash } from '../../common/helpers/generate-hash';
+
+export const metaService = {
+  async createByIndexer(createMetaInput: CreateMetaInput): Promise<Meta | null> {
+    const metaDB = await metaRepository.get(createMetaInput.hash);
+
+    if (metaDB) return;
+
+    const metaTypeDB = plainToClass(Meta, {
```
osipov-mit, line 17: `plainToClass` is deprecated if I'm not wrong. `plaintToInstance` is okay

idea/meta-storage/src/services/meta/meta.service.ts
```diff
@@ -0,0 +1,51 @@
...
+import { getProgramMetadata } from '@gear-js/api';
+
+import { CreateMetaInput } from './input/create-meta.input';
+import { Meta } from '../../database/entities';
+import { metaRepository } from '../../database/repositories';
+import { MetaAlreadyExists, MetadataNotFound } from '../../common/errors';
+import { generateHash } from '../../common/helpers/generate-hash';
+
+export const metaService = {
+  async createByIndexer(createMetaInput: CreateMetaInput): Promise<Meta | null> {
```
osipov-mit, line 12: What does `byIndexer` mean? Name `create` would be clear enough. And you can rename `create` function to `update` because it actually does update an existed record

idea/meta-storage/src/rabbitmq/init-rabbitmq.ts
```diff
@@ -0,0 +1,87 @@
...
+export async function initAMQ(): Promise<void> {
+  connectionAMQP = await connectAMQP(config.rabbitmq.url);
+  mainChannelAMQP = await connectionAMQP.createChannel();
+  const directExchange = RabbitMQExchanges.DIRECT_EX;
+  const directExchangeType = 'direct';
+
+  await mainChannelAMQP.assertExchange(directExchange, directExchangeType);
+
+  await mainChannelAMQP.assertQueue(RMQServices.META_STORAGE, {
+    durable: false,
```
osipov-mit, line 21: Why it isn't durable?

idea/meta-storage/src/rabbitmq/init-rabbitmq.ts
```diff
@@ -0,0 +1,87 @@
...
+    );
+  } catch (error) {
+    console.error(`${new Date()} | Direct exchange consumer error`, error);
+  }
+}
+
+async function handleIncomingMsg(method: string, params: any): Promise<any> {
+  const methods = {
+    [API_METHODS.META_ADD]: () => {
+      if ('hash' in params) {
```
osipov-mit, line 63: Every request will invovle `hash` prop. What is the meaning of this condition?

idea/meta-storage/src/services/meta/meta.service.ts
osipov-mit: And actually it's better to keep just one method

idea/meta-storage/src/database/repositories/meta.repository.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Meta } from '../entities/';
+
+const metaRepo = dataSource.getRepository(Meta);
+
+export const metaRepository = {
+  async save(meta: Meta): Promise<Meta> {
+    return metaRepo.save(meta);
+  },
+  async get(id: string): Promise<Meta> {
+    return metaRepo.findOne({ where: { id } });
```
osipov-mit, line 11:
```suggestion
    return metaRepo.findOneBy({ id });
```

idea/meta-storage/src/database/entities/meta.entity.ts
```diff
@@ -0,0 +1,14 @@
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+import { IMeta } from '@gear-js/common';
+
+@Entity()
+export class Meta implements IMeta {
+  @PrimaryColumn()
+  public id!: string; //hash
```
osipov-mit, line 7: Let's name it `hash`

idea/meta-storage/src/database/entities/index.ts
```diff
@@ -0,0 +1 @@
+export * from './meta.entity';
```
osipov-mit, line 1: Is it really needed to create separate folder for just one file?

idea/meta-storage/src/common/helpers/index.ts
osipov-mit, line 1: What is the purpose of this file? 😄

idea/meta-storage/src/common/helpers/generate-hash.ts
```diff
@@ -0,0 +1,6 @@
+import { generateCodeHash } from '@gear-js/api';
+import { HexString } from '@polkadot/util/types';
+
+export function generateHash(hex: HexString): HexString {
+  return generateCodeHash(hex);
+}
```
osipov-mit, line 6: Function that returns another function? It looks like it's redudant

idea/meta-storage/src/common/errors/metadata.ts
```diff
@@ -0,0 +1,9 @@
+import { GearJsonRPCError } from './base';
+
+export class MetadataNotFound extends GearJsonRPCError {
+  name = 'MetadataNotFound';
+}
+
+export class MetaAlreadyExists extends GearJsonRPCError {
+  name = 'MetaAlreadyExists';
+}
```
osipov-mit, line 9: Maybe these errors should be in the `common` package amoung others?

idea/indexer/src/rabbitmq/index.ts
```diff
@@ -100,6 +98,11 @@ export class RMQService {
     this.mainChannel.publish(exchange, queue, Buffer.from(messageBuff), { correlationId });
   }
 
+  static sendMsgToMetaStorage(exchange: RabbitMQExchanges, queue: RabbitMQueues, params: any, method?: string): void {
+    const messageBuff = JSON.stringify(params);
+    mainChannel.publish(exchange, queue, Buffer.from(messageBuff), { headers: { method } });
+  }
+
```
osipov-mit, line 105: Oh, maybe we can do it in some other way? I mean without static method

idea/indexer/src/gear/indexer.ts
```diff
@@ -237,8 +236,15 @@ export class GearIndexer {
       const codeId = id.toHex();
       const metahash = await getMetahash(this.api.code, codeId);
 
+      RMQService.sendMsgToMetaStorage(
+        RabbitMQExchanges.DIRECT_EX,
+        RMQServices.META_STORAGE as any,
+        { hash: metahash },
+        API_METHODS.META_ADD
+      );
+
```
osipov-mit, line 245: I'd find another way to do it. Maybe send msg in TempState class or sth else

osipov-mit (changes requested):

idea/common/src/interfaces/api-requests.ts
```diff
@@ -38,18 +38,14 @@ interface FindProgramParams extends IGenesis, Pick<IProgram, 'id'> {
   owner?: string;
 }
 
-interface AddMetaByCodeParams extends IGenesis, Pick<ICode, 'id'> {
-  metaHex: string;
+interface AddMetaParams extends IGenesis {
```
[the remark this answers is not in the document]
Almazatun: In my opinion, it will be better to have `genesis` in the params, because we already have validation middleware that checks genesis field in the body. Also when `client` for working with `idea` definitely needs to select a network, it's not a big problem if in the request body will exists `genesis` param.

idea/common/src/interfaces/meta.ts
Almazatun: I don't think so, it's not a good idea, because this `interface` have used a `meta` entity.

idea/meta-storage/src/services/meta/meta.service.ts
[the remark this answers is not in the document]
Almazatun: There is a different purpose, `input` it's like `dto` from what kind of data will be created some `entity`. Maybe `IMeta` good idea to use like the `response` type but not a good idea to use like `input`.

idea/meta-storage/src/services/meta/meta.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (metaDB) return;
+
+    const metaTypeDB = plainToClass(Meta, {
+      id: createMetaInput.hash,
+    });
+
+    return metaRepository.save(metaTypeDB);
+  },
+
+  async create(createMetaInput: CreateMetaInput): Promise<Meta> {
```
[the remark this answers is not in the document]
Almazatun: I have separated methods because from where comes the request we should have different behaviors. What I mean by that:
1. every request that comes from `indexer` does not need to call an exception that some `meta` exist or not by hash, just needs to handle and also this request does not need to use a request and reply pattern.
2. when a request comes from `idea` need to call some exception if meta already exists, and of course, we should reply at the end of the process.

idea/meta-storage/src/services/meta/meta.service.ts
Almazatun: https://github.com/gear-tech/gear-js/pull/1282#discussion_r1183954704

idea/meta-storage/src/rabbitmq/init-rabbitmq.ts
Almazatun: Not every, it depends on from where comes a request. Because from `idea` client will send only `hex` by using API method `meta.add`. If a request comes from `indexer` only `hash`.

idea/meta-storage/src/common/errors/metadata.ts
Almazatun: In `common` package we contain only `JSON-RPC-ERROR` not the exceptions. In this service, I have just followed and have made the same as in `indexer` service.

For example - https://github.com/gear-tech/gear-js/blob/main/idea/indexer/src/common/errors/program.ts

idea/meta-storage/src/services/meta/input/create-meta.input.ts
Almazatun: https://github.com/gear-tech/gear-js/pull/1282#discussion_r1183931979

idea/indexer/src/rabbitmq/index.ts
Almazatun: What is the big problem if use this way? Another way to pass a `rabbitMQ` instance is like a constructor parameter in `indexer` and also in `temp-state`.

idea/meta-storage/src/database/entities/index.ts
osipov-mit: For what future? There will be only one entity

idea/common/src/interfaces/meta.ts
Almazatun: Done

idea/tests/src/e2e/e2e.test.ts
```diff
@@ -17,7 +17,7 @@ import {
   getState,
   getStates,
   getStatesByFuncName,
-  mapProgramStates,
+  mapProgramStates, matProgramsMetaHash,
```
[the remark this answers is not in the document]
Almazatun: Renamed. Thanks 🙏.

idea/meta-storage/src/rabbitmq/types/index.ts
```diff
@@ -0,0 +1,8 @@
+interface IRMQMessage {
```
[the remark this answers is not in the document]
Almazatun: Deleted

idea/meta-storage/src/services/meta/meta.service.ts
Almazatun: Done

idea/meta-storage/src/rabbitmq/init-rabbitmq.ts
Almazatun: Changed.

idea/meta-storage/src/database/entities/meta.entity.ts
Almazatun: Done.

idea/meta-storage/src/rabbitmq/init-rabbitmq.ts
osipov-mit: agree

idea/common/src/interfaces/api-requests.ts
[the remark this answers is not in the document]
osipov-mit: But what if I want to get metadata by its hash without any network context?

PR #1288, opened 2023-05-18
Add message `indexer`

Almazatun:
Resolves # .

@reviewer-or-team


squids/message-indexer/src/subsquid/processor.ts
```diff
@@ -0,0 +1,102 @@
...
+      const { header: { timestamp, hash } } = block;
+      const time = new Date(timestamp);
+
+      if (name === SubSquidMessageEventType.GEAR_USER_MESS_READ) {
+        const { id, reason } = item.event.args;
+        const message = await ctx.store.findOneBy(Message, { id });
+
+        if (message) {
+          // eslint-disable-next-line no-underscore-dangle
+          message.readReason = reason.value.__king;
```
osipov-mit, line 28: Are you sure about the spelling of `__king`?

squids/message-indexer/src/subsquid/processor.ts
```diff
@@ -0,0 +1,102 @@
...
+      // eslint-disable-next-line no-continue
+      if (item.kind !== 'event') continue;
+
+      const { name } = item.event;
+      const { header: { timestamp, hash } } = block;
+      const time = new Date(timestamp);
+
+      if (name === SubSquidMessageEventType.GEAR_USER_MESS_READ) {
+        const { id, reason } = item.event.args;
+        const message = await ctx.store.findOneBy(Message, { id });
```
osipov-mit, line 24: It's better to use a temporary state to store data from each batch and save it into database only once for each batch. It will reduce number of r/w operations to the database as well as increase speed of network indexing.

osipov-mit (changes requested):

squids/message-indexer/src/subsquid/processor.ts
Almazatun: Premature optimization may not good idea sometimes. Currently in the `processor` for the `write option` in a database for each batch elements going only once. I didn't see any points right now to optimize if we don't have any information for that case.

squids/message-indexer/src/subsquid/init-processor.ts
```diff
@@ -0,0 +1,16 @@
+import { SubstrateBatchProcessor } from '@subsquid/substrate-processor';
+
+import config from '../config/configuration';
+import { EventType } from '../common/enum';
+
+const { subsquid } = config;
+
+export const processor = new SubstrateBatchProcessor()
+  .setDataSource({ archive: subsquid.archiveProvider })
+  .addEvent(EventType.GEAR_MSG_ENQUEUED, { data: { event: { args: true, extrinsic: true } } })
```
osipov-mit, line 10:
```suggestion
  .addEvent(EventType.GEAR_MSG_QUEUED, { data: { event: { args: true, extrinsic: true } } })
```

osipov-mit (changes requested):

squids/message-indexer/src/subsquid/processor.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  const messages = Array.from(messageMap.values());
+  if (messages.length >= 1) {
+    await ctx.store.save(messages);
+    messageMap.clear();
+  }
+}
+
+export async function run() {
```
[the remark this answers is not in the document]
osipov-mit: `async` keyword is unnecessary here

squids/message-indexer/src/subsquid/processor.ts
osipov-mit: Of course it appears. I described the problem in the next sentence

> Because in most cases MessageQueued and MessagesDispatched events occur in one block so message that you're trying to get from database in line 40 won't exist.

osipov-mit (changes requested):

squids/message-indexer/src/subsquid/processor.ts
```diff
@@ -21,27 +21,37 @@ async function eventHandler(ctx: BatchContext<Store, BatchProcessorItem<typeof p
...
-          messageMap.set(message.id, message as Message);
+          messageMap.set(updateMsg.id, updateMsg as Message);
         }
       }
 
       if (name === EventType.GEAR_MSG_DISPATCHED) {
         const { statuses } = item.event.args;
-        const [id, status] = Object.entries(statuses) as [string, any];
-        const statusVal = status as any;
+        const [id, status] = statuses[0];
```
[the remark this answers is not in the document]
Almazatun: I can add a new `field` to the `message` entity if need to save all `DispatchStatuses `. Does it really need it @osipov-mit ?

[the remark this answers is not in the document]
osipov-mit: You only take here the first status, but what about the others? I mean if there is more than one status

PR #1345, opened 2023-08-04
Migrate to meta storage service

nikitayutanov:


idea/frontend/src/hooks/useProgramActions/useProgramActions.tsx
```diff
@@ -132,7 +134,16 @@ const useProgramActions = () => {
       if (resolve) resolve();
 
       if (isDevChain) {
-        await uploadLocalProgram({ id: programId, name: name || programId, owner: account?.decodedAddress! });
+        const metahash = await api.code.metaHash(codeId);
```
osipov-mit, line 137:
```suggestion
        const metahash = await api.program.metaHash(programId);
```
I guess it should work faster.

PR #1370, opened 2023-09-01
Fix the problem of inconsistency between sendMessage parameters and t…

harodggg:
…he chain

Resolves # .
When using gear sdk, I encountered the following error:
image

After research, I found out that the sendmessage of gear.js has one more prepaid parameter than the chain：
image
image

Therefore, I deleted the parameter prepaid of the sendmessage function of the gear.js sdk to ensure that the interface on the chain is consistent with that of the gear.js sdk to solve this problem.

@reviewer-or-team


osipov-mit:
@harodggg you can use previous version of `gear-js/api` which hasn't this parameter. `0.33.0` version supports newer version of gear runtime and this parameter is mandatory from that version

PR #1372, opened 2023-09-04
Improve explorer decoding

nikitayutanov:
- Fix log events payolad decoding
- Add payload decoding for `gear.sendMessage`, `gear.sendReply`, `gear.uploadProgram`, `gear.createProgram` extrinsics
- Move explorer entity to `features` folder


idea/frontend/src/features/explorer/ui/decoded-preformatted-block/decoded-preformatted-block.tsx
```diff
@@ -0,0 +1,160 @@
...
+      default:
+    }
+  };
+
+  const getFallbackType = () => {
+    if (!metadata) return;
+
+    switch (method) {
+      case Method.UserMessageSent:
+        return metadata.types.init.output;
```
osipov-mit, line 89: Actually there can be types.other.output here as well

PR #1451, opened 2023-11-09
Fix `api.getExtrinsicFailed` method

osipov-mit:
## Changes
- Fix `api.getExtrinsicFailedError` method. It now returns an object with the formatted docs, method and name of the error

### DIFF
--- a/api/src/GearApi.ts
+++ b/api/src/GearApi.ts
@@ -157,13 +157,15 @@ export class GearApi extends ApiPromise {
         method: data?.method,
         name: data?.name,
       };
-    } else if (error.isCannotLookup) {
+    }
+    if (error.isCannotLookup) {
       return {
         docs: null,
         method: 'CannotLookup',
         name: 'CannotLookup',
       };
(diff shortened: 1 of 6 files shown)


api/src/GearApi.ts
```diff
@@ -149,9 +148,33 @@ export class GearApi extends ApiPromise {
...
-    const { isModule, asModule } = error;
-    return isModule ? this.registry.findMetaError(asModule) : null;
+    if (error.isModule) {
+      const data = this.registry.findMetaError(error.asModule);
+      return {
+        docs: data?.docs.join(' '),
+        method: data?.method,
+        name: data?.name,
+      };
+    } else if (error.isCannotLookup) {
```
ElessarST, line 160: it seems that we can remove else here and keep only ifs. What do you think?

osipov-mit: Right! Thanks

ElessarST (approved):

EugenWay (approved):

nikitayutanov (approved):

PR #1458, opened 2023-11-28
Add VaraApiV1000 and VaraApiV1010 classes

osipov-mit:


api/src/Message.ts
```diff
@@ -212,7 +225,7 @@ export class GearMessage extends GearTransaction {
    * ```
    */
   sendReply(
-    args: IMessageSendReplyOptions | VaraMessageSendReplyOptions,
+    args: V1000MessageSendReplyOptions | V1010MessageSendReplyOptions,
```
ElessarST, line 228: maybe create a new type `type MessageSendReplyOptions = V1000MessageSendReplyOptions | V1010MessageSendReplyOptions`?

api/src/Storage.ts
```diff
@@ -48,11 +48,11 @@ export class GearProgramStorage {
   async getProgramPages(programId: HexString, program: GearCommonActiveProgram, at?: HexString): Promise<IGearPages> {
     const pages = {};
     const query =
-      this._api.specVersion >= 1020
+      this._api.specVersion >= 1030
```
ElessarST, line 51: nit: `1030` can be moved to configs or const

api/src/Message.ts
osipov-mit: done

api/src/Storage.ts
osipov-mit: Yes, thanks

ElessarST (approved):

faizov (approved):

PR #1515, opened 2024-03-22
feat(idea): introduce indexer for vouchers

osipov-mit:


idea/voucher-indexer/.env
```diff
@@ -0,0 +1,7 @@
+DB_NAME=voucher_indexer
```
nikitayutanov, line 1: Should be `env.example` I guess?

nikitayutanov (approved):

PR #1613, opened 2024-08-10
feat(meta-storage) change RMQ to REST

Zewasik:
Resolves # .

- remove RMQ service
- add 4 endpoints for handling meta and sails request
- add tests for http server and meta service
- add swagger documentation

@osipov-mit


idea/meta-storage-2/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "allowSyntheticDefaultImports": true,
+    "strict": false
+  }
```
osipov-mit, line 9: What do you think about adding `tsconfig.build.json` file having `"include": ["src"]` prop for compiling? This way we won't have `tests` dir built during compilation

idea/meta-storage-2/src/meta.router.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  constructor(private metaService: MetaService) {}
+
+  public init(): void {
+    this.router.use(express.json());
+    this.router
+      .get('/meta', this.getMetaDetails.bind(this))
+      .post('/meta', this.addMetaDetails.bind(this))
+      .get('/sails', this.getSails.bind(this))
+      .post('/sails', this.addSails.bind(this));
```
osipov-mit, line 33: Is `.bind` necessary here?

idea/meta-storage-2/src/server.ts
```diff
@@ -0,0 +1,25 @@
...
+  .then(main)
+  .then((app) => {
+    app.listen(port, () => {
+      logger.info(`Server is running on port ${port}`);
+    });
+  })
+  .catch((error) => {
+    logger.error("Can't start the indexer", { error, stack: error.stack });
+    process.exit(1);
+  });
```
osipov-mit, line 25: I'd suggest to move all this stuff to `main.ts` and initialise all services in `main` function

idea/meta-storage-2/src/service.ts
```diff
@@ -0,0 +1,120 @@
...
+      throw new MetaNotFoundError();
+    }
+
+    return meta;
+  }
+
+  async getAllWithState(): Promise<string[]> {
+    const meta = await this.metaRepo.find({ where: { hasState: true }, select: { hash: true } });
+    return meta.map((m) => m.hash);
+  }
```
osipov-mit, line 86: Is this method used somewhere?

idea/meta-storage-2/src/server.ts
Zewasik: In this case, I won't be able to properly friend the services due to the lack of dependency injection capability. Alternatively, I could rename the main file to avoid confusion

idea/meta-storage-2/src/service.ts
osipov-mit: Ah, I see, I think it can be removed

idea/meta-storage-2/src/meta.router.ts
Zewasik: Yes, without the binding the express will not work properly

idea/meta-storage-2/tsconfig.json
Zewasik: In this case, I will exclude the test folder.

idea/meta-storage-2/src/server.ts
osipov-mit: Yes let's rename then

osipov-mit (approved):

PR #1638, opened 2024-09-03
fix(idea/squijd): saving init message payload and setting ProgramSet status

osipov-mit:


idea/squid/src/temp-state.ts
```diff
@@ -14,6 +14,7 @@ import {
   MessageFromProgram,
   MessageToProgram,
   Program,
+  MessageEntryPoint,
```
Zewasik, line 17: looks like the getCtorNamePrefix and MessageEntryPoint imports are not used

Zewasik (changes requested):

Zewasik (approved):

PR #1648, opened 2024-10-01
feat(wallet-connect): init package

nikitayutanov:
Create `@gear-js/wallet-connect` library for a standardized wallet connection UI across applications.


vraja-nayaka (approved):

ereburg (approved):

ereburg:
Maybe better propname for choosing ui styles "gear" or "vara" would be _theme_ instead of _variant_?
"Variant" for me is more about visual design of component or button, eg.: outline, inline, flat, ghost, etc. And, yes, those variants should be standardized across ui design systems.

nikitayutanov:
> Maybe better propname for choosing ui styles "gear" or "vara" would be _theme_ instead of _variant_? "Variant" for me is more about visual design of component or button, eg.: outline, inline, flat, ghost, etc. And, yes, those variants should be standardized across ui design systems.

I was considering `theme` as a prop name and decided not to go with it, cuz `theme` sometimes is used in the context of `dark`/`light` mode, therefore can be confusing. No reason to believe that we will have it one day, but who knows, `vara` in `dark` mode looks promising.

I see your point, however `variant` there was intended to be interpreted as a simplified version of `uiVariant`.

What makes more sense to you:
`variant` to define UI kit and `theme` to define `dark`/`light` mode
or `theme` to define UI kit and `mode` to define `dark`/`light` mode?

@vraja-nayaka would like to know your opinion too.

ereburg:
> What makes more sense to you:
> 
> `variant` to define UI kit and `theme` to define `dark`/`light` mode
> 
> or `theme` to define UI kit and `mode` to define `dark`/`light` mode?


As for me, it is much more convenient to have certain namespaces like vara/gear to use it in our themes (as it is) and have some variations/modes of those themes like dark/light/pink/etc.

nikitayutanov:
> As for me, it is much more convenient to have certain namespaces like vara/gear to use it in our themes (as it is) and have some variations/modes of those themes like dark/light/pink/etc.

Okay, will make a change in the next PR.

PR #1658, opened 2024-10-25
feat(api) add support for `gearEthBridge`

### LINKED ISSUE #1655: feat(api): support `geatEthBridge` pallet
### Problem to Solve

- Support the queries of the pallet
- Support the extrinsics of the pallet

### Possible Solution

.

### Notes

_No response_

Zewasik:
Resolves #1655 .

- add queries and extrinsic types

@osipov-mit

### DIFF
--- a/api/src/EthBridge.ts
+++ b/api/src/EthBridge.ts
@@ -1,0 +1,5 @@
+import { GearApi } from './GearApi';
+
+export class GearEthBridge {
+  constructor(private api: GearApi) {}
+}
--- a/api/src/types/augment/augment-api-query.ts
+++ b/api/src/types/augment/augment-api-query.ts
@@ -2,7 +2,20 @@
 import '@polkadot/api-base/types/storage';
 
 import type { ApiTypes, AugmentedQuery, QueryableStorageEntry } from '@polkadot/api-base/types';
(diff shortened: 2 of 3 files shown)


api/src/EthBridge.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  async authoritySetHash(): Promise<HexString> {
+    const result = await this.api.query.gearEthBridge.authoritySetHash();
+
+    try {
+      const hash = result.unwrap();
+      return hash.toHex();
+    } catch (err) {
+      throw new AuthoritySetHashError(err);
+    }
```
osipov-mit, line 18: There should be methods `.is_some` and `.is_none` if I'm not mistaken. Then it can look like this
```javascript
if (result.is_none()) {
  throw new AuthroitySetHashNotFound(); // cuz if it's `none` it literally not found
}
return result.unwrap().toHex()
```

api/src/EthBridge.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  async clearTimer(): Promise<number> {
+    const result = await this.api.query.gearEthBridge.clearTimer();
+    try {
+      return result.unwrap().toNumber();
```
osipov-mit, line 24: The same as above. Better to check if the value is `Some` before calling unwrap

osipov-mit (changes requested):

osipov-mit (approved):

PR #1663, opened 2024-11-06
feat(idea/voucher-indexer) add vouchers issued for all programs param

### LINKED ISSUE #1630: feat(idea/voucher-indexer): get program vouchers for spender, is program voucher exists requests
### Problem to Solve

Based on #1629.

Send message page, there is program vouchers select for a specific account.
Programs page, there is voucher badge in a program card that demonstrates if program has any active vouchers.

 We can use existing `/vouchers` request to obtain data required for them, however there are few limitations:
- List in a response is paginated, only first `n` vouchers will be displayed by default. Since it's a select element, we can't really add pagination there, from a frontend perspective we can only change the design or specify higher limit.
- Need to send two requests at the same time to display specific program vouchers and vouchers issued for all programs - one with `programs: [programId]` argument, and another one with `programs: []` argument.
- Not sure about this one, but feels like `programs` property in a reponse is always empty array, regardless if voucher is for programs and codes or codes only.

### Possible Solution

Ideally, we can add two new requests to handle these use cases, or enhance the existing one.

Zewasik:
Resolves #1630 .

- add param for vouchers to get vouchers with empty programs

@osipov-mit

### DIFF
--- a/idea/explorer/src/services/voucher.ts
+++ b/idea/explorer/src/services/voucher.ts
@@ -68,7 +68,6 @@ export class VoucherService {
       }
 
       qb.andWhere(`(${conditions.join(' OR ')})`, params);
-      console.log(qb.getQuery());
     }
 
     if (expired !== undefined) {
--- a/idea/explorer/src/services/voucher.ts
+++ b/idea/explorer/src/services/voucher.ts
@@ -34,6 +34,7 @@ export class VoucherService {
     limit,
     offset,
     expired,
+    includeAllPrograms,
   }: ParamGetVouchers): Promise<ResManyResult<Voucher>> {
     const qb = this._repo.createQueryBuilder('v');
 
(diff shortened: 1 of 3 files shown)


idea/explorer/src/services/voucher.ts
```diff
@@ -53,15 +54,20 @@ export class VoucherService {
...
+          return `v.programs::jsonb ? :p${i}`;
+        }) || [];
+
+      if (includeAllPrograms) {
+        conditions.push(`jsonb_array_length(v.programs) = 0`);
       }
-      qb.andWhere(where + ')', params);
+
+      qb.andWhere(`(${conditions.join(' OR ')})`, params);
       console.log(qb.getQuery());
```
osipov-mit, line 71: Let's remove this console.log

osipov-mit (approved):

PR #1677, opened 2024-11-28
chore(js): bump dependencies

osipov-mit:


nikitayutanov (approved):

nikitayutanov:
Why `Buffer` type is not supported anymore?

vraja-nayaka (approved):

Zewasik (approved):

osipov-mit:
> Why `Buffer` type is not supported anymore?

There is no need in it actually since it was used only for code uploading from nodejs. And polkadot pkgs don’t support it so it’s quite inconvenient to support it on our side.

PR #1690, opened 2024-12-03
feat(vara-ui): upd checkbox

ereburg:
- upd light theme according to new vara styles;
- add dark theme;
- upd stories and types.

### DIFF
--- a/utils/vara-ui/src/components/checkbox/checkbox.module.scss
+++ b/utils/vara-ui/src/components/checkbox/checkbox.module.scss
@@ -1,36 +1,198 @@
 .label {
   display: flex;
   align-items: center;
+  gap: 8px;
   cursor: pointer;
+  user-select: none;
 
-  &.disabled {
-    pointer-events: none;
-    opacity: 0.3;
+  &:has(.checkbox) {
+    position: relative;
+
+    &::after {
+      content: '';
+      position: absolute;
+      top: 3px;
+      left: 3px;
+      width: var(--checkbox-size);
+      height: var(--checkbox-size);
+      border-color: #00ffc466;
+      background-color: #000000;
+      mask-image: url("data:image/svg+xml,%3Csvg width='18' height='18' viewBox='0 0 18 18' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M2 9.00008L7 14.0001L16 5.00008L14.59 3.58008L7 11.1701L3.41 7.59008L2 9.00008Z' fill='black'/%3E%3C/svg%3E%0A");
+      opacity: 0;
+      transition-duration: 250ms;
+      transition-property: background-color, opacity;
+      transition-timing-function: ease;
+    }
+
+    &:has(:checked) {
+      &::after {
+        opacity: 1;
+      }
+
+      &:has(:disabled) {
+        &::after {
+          background-color: rgba(208, 213, 221, 1);
+        }
+      }
+    }
   }
+
+  //Show disabled state even if an error displayed
+  &[aria-invalid=true]:not(.disabled) {
+    .checkbox {
+      border-color: rgba(255, 51, 51, 0.8);
+
+      &:checked {
+        background-color: rgba(255, 51, 51, 0.8);
+      }
+    }
+
+    &:has(.checkbox) {
+      &::after {
+        background-color: #ffffff !important;
+      }
+    }
+  }
+}
+
+.disabled {
+  cursor: not-allowed;
 }
 
 .input {
   appearance: none;
   cursor: inherit;
-  margin: 0 10px 0 0;
+  margin: 0;
+
+  outline: none;
+  outline-offset: 2px;
+
+  &:disabled {
+    cursor: not-allowed;
+  }
+
+  &:focus-visible {
+    outline: rgba(0, 255, 196, 1) 2px solid;
+  }
 }
 
+// Types
 .checkbox {
-  width: 15px;
-  height: 15px;
-  background-image: url("data:image/svg+xml,%3Csvg width='15' height='15' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Crect x='.5' y='.5' width='14' height='14' rx='1.5' stroke='%23909090'/%3E%3C/svg%3E");
+  flex-basis: var(--checkbox-size);
+  width: var(--checkbox-size);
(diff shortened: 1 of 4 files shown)


utils/vara-ui/src/components/checkbox/checkbox.tsx
```diff
@@ -1,28 +1,36 @@
...
 import cx from 'clsx';
 import styles from './checkbox.module.scss';
+import type { ICheckboxSizes, ICheckboxTypes } from './helpers.ts';
 
 type Props = InputHTMLAttributes<HTMLInputElement> & {
   label: string;
-  type?: 'switch';
+  type?: ICheckboxTypes;
+  checkboxSize?: ICheckboxSizes;
+  hasError?: boolean;
```
nikitayutanov, line 10:
```suggestion
  type?: 'checkbox' | 'switch';
  size?: 'default' | 'small';
  error?: string;
```

No need for extra file to handle types, it's small enough to be inlined, and let's stick to existing naming conventions

utils/vara-ui/src/components/checkbox/checkbox.module.scss
```diff
@@ -1,36 +1,198 @@
 .label {
   display: flex;
   align-items: center;
+  gap: 8px;
   cursor: pointer;
+  user-select: none;
```
nikitayutanov, line 6: I understand that `user-select: none` probably prevents text from being selected on double click, but I think it can significantly affects UX. Sometimes there really can be a need to copy paste label's text.

utils/vara-ui/src/components/checkbox/checkbox.tsx
```diff
@@ -1,28 +1,36 @@
...
-        ref={ref}
-        {...attrs}
-      />
+    return (
+      <label
+        className={cx(
+          styles.label,
+          className,
+          hasError && styles.error,
+          disabled && styles.disabled,
```
nikitayutanov, line 23: `hasError && styles.error` and `disabled && styles.disabled` may be not needed, since label styles are controlled via `:has` selector?

utils/vara-ui/src/components/checkbox/checkbox.module.scss
```diff
@@ -1,36 +1,198 @@
...
+    &:has(.checkbox) {
+      &::after {
+        background-color: #ffffff !important;
+      }
+    }
   }
 }
 
+.disabled {
+  cursor: not-allowed;
```
nikitayutanov, line 59: I think it's an overkill to use `not-allowed` cursor on the disabled input. Default cursor is `pointer`, and it already indicates that an element is interactive, so it might be more appropriate to use the default cursor for disabled inputs.

utils/vara-ui/src/components/checkbox/checkbox.module.scss
```diff
@@ -1,36 +1,198 @@
...
+
+  //Show disabled state even if an error displayed
+  &[aria-invalid=true]:not(.disabled) {
+    .checkbox {
+      border-color: rgba(255, 51, 51, 0.8);
+
+      &:checked {
+        background-color: rgba(255, 51, 51, 0.8);
+      }
+    }
```
nikitayutanov, line 48: If we're targeting checkbox there, and not label, better to nest it in a `.checkbox` class itself, otherwise it's quite hard to read and follow the logic.

utils/vara-ui/src/components/checkbox/checkbox.module.scss
```diff
@@ -1,36 +1,198 @@
...
   display: flex;
   align-items: center;
+  gap: 8px;
   cursor: pointer;
+  user-select: none;
 
-  &.disabled {
-    pointer-events: none;
-    opacity: 0.3;
+  &:has(.checkbox) {
```
nikitayutanov, line 8: Isn't `&:has(.checkbox)` is a redudant here? `.label` always `has` checkbox, am I missing something?

utils/vara-ui/src/components/checkbox/checkbox.module.scss
```diff
@@ -1,36 +1,198 @@
...
+      border-color: rgba(255, 51, 51, 0.8);
+
+      &:checked {
+        background-color: rgba(255, 51, 51, 0.8);
+      }
+    }
+
+    &:has(.checkbox) {
+      &::after {
+        background-color: #ffffff !important;
```
nikitayutanov, line 52: For me it's working without `!important`, am I missing something?

utils/vara-ui/src/components/checkbox/checkbox.module.scss
```diff
@@ -1,36 +1,198 @@
...
+  .checkbox {
+    margin: 2.5px;
+  }
+}
+
+:global(.dark-theme) {
+  .label {
+    color: #ffffff;
+
+    &:has(.checkbox) {
```
nikitayutanov, line 161: Same thing above `&:has(.checkbox)` as above.

utils/vara-ui/src/components/checkbox/checkbox.module.scss
```diff
@@ -1,36 +1,198 @@
...
+
+  .checkbox {
+    background-color: rgba(255, 255, 255, 0.03);
+    border-color: rgba(255, 255, 255, 0.04);
+
+    &:hover {
+      border-color: rgba(0, 255, 196, 0.4);
+    }
+
+    &:focus-visible {
```
nikitayutanov, line 180: What's the reason to use `focus-visible` instead of default `focus`?

utils/vara-ui/src/components/checkbox/checkbox.module.scss
```diff
@@ -1,36 +1,198 @@
...
+
+  outline: none;
+  outline-offset: 2px;
+
+  &:disabled {
+    cursor: not-allowed;
+  }
+
+  &:focus-visible {
+    outline: rgba(0, 255, 196, 1) 2px solid;
```
nikitayutanov, line 75: `outline` on focus is not per Figma template, as I can see.

nikitayutanov:
Dark mode checkbox is not working as expected on present `disabled` or `error` states:

image
image

It's hover state on error is wrong too:
image

nikitayutanov:
`small` checkbox is slightly off:
image

vraja-nayaka (approved):

PR #1692, opened 2024-12-04
feat(vara-ui): add `danger` color, spinner loader and dark mode to `Button`

ereburg:
- upd component styles for new vara ui
- add dark theme support


utils/vara-ui/src/components/button/button.tsx
```diff
@@ -1,12 +1,13 @@
 import { ButtonHTMLAttributes, FunctionComponent, ReactNode, SVGProps, forwardRef } from 'react';
 import cx from 'clsx';
 import styles from './button.module.scss';
+import type { IButtonColors, IButtonSizes } from './helpers.ts';
 
 type BaseProps = ButtonHTMLAttributes<HTMLButtonElement> & {
   text?: string;
   icon?: FunctionComponent<SVGProps<SVGSVGElement> & { title?: string | undefined }>;
-  color?: 'primary' | 'dark' | 'light' | 'grey' | 'border' | 'transparent';
```
vraja-nayaka, line 8: There are no more "dark" and "light" color names in the design system.

Since the introduction of the dark theme, these names have become misleading: when working with the dark theme, a "dark" button will appear light.

We can keep them marked as "@deprecated" and add an "additional" color (which matches "dark" in terms of styles), as seen in the design.

Ideally, once the migration is complete we can remove them entirely.
There aren’t many places using them in dapps

nikitayutanov: Extremely good point!

utils/vara-ui/src/components/button/helpers.ts
```diff
@@ -0,0 +1,8 @@
+const buttonSizes = ['x-small', 'small', 'medium', 'default', 'x-large'] as const;
+type IButtonSizes = (typeof buttonSizes)[number];
```
nikitayutanov, line 2: We're not following `I` naming convention for types, and at the same time it's not an interface.

Also, I think separate file for small types like that is an overkill.

utils/vara-ui/src/components/button/helpers.ts
```diff
@@ -0,0 +1,8 @@
...
+
+const buttonColors = ['primary', 'dark', 'light', 'grey', 'border', 'transparent', 'destructive'] as const;
```
nikitayutanov, line 4: `error` or `warning` more descriptive than `destructive`

utils/vara-ui/src/components/button/button.module.scss
```diff
@@ -1,46 +1,44 @@
...
+  transition-property: color, background-color, box-shadow;
+  transition-duration: 250ms;
+
+  cursor: pointer;
+  outline: none;
+  outline-offset: 2px;
 
   &:disabled {
-    pointer-events: none;
+    cursor: not-allowed;
```
nikitayutanov, line 23: `cursor: not-allowed` on a disabled element with `cursor: pointer` is an UX overkill.

utils/vara-ui/src/components/button/button.tsx
```diff
@@ -52,7 +53,7 @@ const Button = forwardRef<HTMLButtonElement, Props>((props, ref) => {
       className={cx(
         styles.button,
         styles[color],
-        color !== 'transparent' && styles[size],
+        styles[size],
```
nikitayutanov, line 56: We can't remove transparent button without size, it can break UI in many places.

utils/vara-ui/src/components/button/button.module.scss
```diff
@@ -1,46 +1,44 @@
...
-  &.disabled {
-    opacity: 0.3;
+  &:focus-visible {
+    outline: rgba(0, 255, 196, 1) 2px solid;
   }
 
-  &.loading {
-    animation: blink 1s linear infinite forwards;
-    background-size: 200%;
+  svg {
```
nikitayutanov, line 30: I suggest to use class there and not general svg selector, to only target icons passed via `SVG` prop.

There can be cases when svg is passed via `children` prop, Identicon for example, for which size should not apply.

Also, I don't think that properties other than fixed width and height are useful there, for a size they are superior and should be enough. For color, we can't do much since `path` element is usually colored but it's not guaranteed, and also, some of the icons can have different color palette in terms of shades that we should't change.

utils/vara-ui/src/components/button/helpers.ts
vraja-nayaka: Most common is `danger`

vraja-nayaka (approved):

PR #1694, opened 2024-12-04
feat(vara-ui): add `large` size and dark mode support to `Input`, drop floating label

ereburg:


vraja-nayaka (approved):

utils/vara-ui/src/components/input/input.module.scss
```diff
@@ -127,34 +158,97 @@
   bottom: 0;
   left: 0;
   right: 0;
+  z-index: -1;
 
-  border: 1px solid #0000003b;
   border-radius: 4px;
+  background-color: #ffffff;
```
nikitayutanov, line 164: We can't put white background color as we can't guarantee that every application's background will be white too. It should be transparent.

nikitayutanov:
Another comments from previous PRs about UX and our style practices applies there too.

utils/vara-ui/src/components/input/input.tsx
```diff
@@ -63,6 +64,7 @@ const Input = forwardRef<HTMLInputElement, Props>(
             className={cx(styles.input, styles[size], error && styles.error)}
             placeholder={placeholder}
             ref={ref}
+            disabled={disabled}
```
nikitayutanov, line 67: `disabled` is applied from `attrs` destructuring already.

nikitayutanov:
Another comments from previous PRs about UX and our style practices applies there too.

nikitayutanov:
Also, regarding general approach to dark mode styles.

What do you think about creating mixin, that will abstract dark mode for a certain element?
I'm not sure that it will work, but something like:

```
@mixin darkMode ($width) {
  :global(.dark-mode) & {
    @content;
  }
}
```

to utilize it like that:

```
.element {
  color: #fff;

  @include darkMode {
    color: #000;
  }
}
```

If it's working as expected, I believe it would be far easier to control color variations in the scope of a certain element.
`Input` styles are bright example of it, since some properties can be easilly forgotten or missed.

vraja-nayaka (approved):

PR #1696, opened 2024-12-04
feat(vara-ui): add dark mode to `Select`, drop floating label

ereburg:


utils/vara-ui/src/components/select/select.module.scss
```diff
@@ -138,3 +166,72 @@
 .block {
   width: 100%;
 }
+
+:global(.dark-theme) {
```
vraja-nayaka, line 170: Seems like we need styles for options in dark-theme
image

nikitayutanov:
Same comments as in Input PR.

PR #1701, opened 2024-12-09
feat(idea/frontend) add query params

### LINKED ISSUE #1691: feat(idea/frontend) add location query and history
### Problem to Solve

On the program and message page, it's hard to navigate back to the old state, because it resets to the initial one when you go back to the page. Additionally, on the same pages and voucher page, when you share and reload the page, you can't get the same state that was on that page. For example, when selecting a program message and returning to the list of messages, all filters are reset and need to be filled in again.

### Possible Solution

Have the ability to conveniently navigate between tabs and pages, keeping all filters and selected tabs intact. For example, this can be done by using query parameters and adding individual records to the histories.

### Notes

_No response_

Zewasik:
Resolves #1691 .

- add query params for programs, messages, vouchers and mailbox

@osipov-mit

### DIFF
--- a/idea/frontend/src/features/message/ui/program-messages/program-messages.tsx
+++ b/idea/frontend/src/features/message/ui/program-messages/program-messages.tsx
@@ -10,6 +10,7 @@ import { SailsFilterGroup } from '@/features/sails';
 
 import { useMessagesToProgram, useMessagesFromProgram } from '../../api';
 import { MessageCard } from '../message-card';
+import { useSearchParams } from 'react-router-dom';
 
 type Props = {
   programId: HexString;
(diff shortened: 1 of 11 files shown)


osipov-mit:
@Zewasik pls update branch
@nikitayutanov pls review

nikitayutanov:
Good solution so far, but I think the task is a little bit more complicated than it seems - I guess we need some generic solution, rather than implementing it over in each component that has filters functionality, otherwise it's gonna be hard to maintain.

The way I see it:

- We have to handle default filter state values - merge default empty values with the ones from search params.
Perhaps, it should be possible to define one generic hook for it - `useFilterValues` as a part of `filters` feature.
It would accept default values as a parameter and perform necessary logic of combining it with URL params under the hood.

- We have to set URL params on filter value submit. I believe would make sense to embed it's logic directly onto submit handlers - either common one in `Filters` component itself, or to each separate component such as filter `Radio`/`StatusRadio`/`StatusCheckbox`. This way we can eliminate the need to use side effects inside of `useEffect`, and generalize URL params set in one place to reuse everywhere.

- Specified above principles should also apply to `SearchForm` component as well.

Apart from it, here's concerns to be aware of even in the current implementation:

- Reset filters button should reset filter to it's empty state, even if URL params are present. To reproduce - open URL with defined params -> reset button will appear only on changed value, but should already be present, since custom value is provided from URL.
 
- Some kind of check should be performed to determine whether provided value from URL is a valid one. To reproduce - open URL with defined params -> change param to a random value, that is not the one present in filters.

PR #1707, opened 2024-12-13
feat(idea/frontend): add code verifier

nikitayutanov:


ereburg (approved):

vraja-nayaka (approved):

idea/gear/frontend/src/features/code-verifier/api/hooks.ts
```diff
@@ -0,0 +1,54 @@
...
+
+function useVerifyCode() {
+  return useMutation({
+    mutationKey: ['verify-code'],
+    mutationFn: verifyCode,
+  });
+}
+
+function useIsCodeVerified(codeId: HexString | null | undefined) {
+  const alert = useAlert();
```
osipov-mit, line 17: I think it can be worth to have some kind of cache for verified code ids just to avoid making unnecessary requests

nikitayutanov: We're using `@tanstack/react-query` lib to handle async states, caching requests is part of it's functionality.

osipov-mit (approved):
Just run `bump:gear-idea minor` to update idea on stage-env

PR #1708, opened 2024-12-18
fix(vara-ui): add `Anuphan` font fallback

nikitayutanov:


utils/vara-ui/src/assets/styles/index.css
```diff
@@ -1,35 +1,6 @@
...
-
-*,
-::after,
-::before {
-  box-sizing: border-box;
-}
+@import url('./resets.css');
 
 body {
   font-family: 'Geist Variable', sans-serif;
```
ereburg, line 5: It's better to use Anuphan font value directly, because there is already a fallback value - sans-serif. Previous declaration wont be ever reached. 

`font-family: 'Geist Variable', Anuphan, sans-serif;`

nikitayutanov: The goal is not to add fallback font literally, but rather the ability to migrate flawlessly from previous versions.

Import as it is right now - `import '@gear-js/vara-ui/dist/style.css'` will utilize `Geist`,
in case `Anuphan` font is needed - `import '@gear-js/vara-ui/dist/style-deprecated.css'` should be specified.

There won't be any overrides since font style are in completely separate files.

ereburg (approved):

PR #1712, opened 2025-01-13
feat(utils) add general logger

Zewasik:
Resolves # .

- add new module "gear-logger"

@osipov-mit


utils/gear-logger/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
```
osipov-mit, line 1: I think we either need types or don't need typescript at all

osipov-mit: Yes, I mean that if don't build types than we don't need typescript here at all. This package can be just a js package. So is the `logger` variable exported from this package typed?

package.json
```diff
@@ -31,7 +31,8 @@
     "utils/gear-hooks",
     "utils/gear-ui",
     "utils/vara-ui",
-    "utils/wallet-connect"
+    "utils/wallet-connect",
+    "utils/gear-logger"
```
osipov-mit, line 35: Let's rename this directory to just `logger`

utils/gear-logger/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "gear-js"
+  ],
+  "scripts": {
+    "build": "rm -rf dist && tsc"
+  },
+  "dependencies": {
+    "winston": "3.7.2"
+  },
+  "devDependencies": {
+    "typescript": "4.7.4"
```
osipov-mit, line 23: not necessary since it's declared in the root package.json

utils/gear-logger/package.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "type": "module"
```
osipov-mit, line 25: The [`main`, `module`, `types`] or `exports` fields are mandatory here for proper import into other packages

utils/gear-logger/src/index.ts
```diff
@@ -0,0 +1,2 @@
+import { logger } from './logger';
+export { logger };
```
osipov-mit, line 2:
```suggestion
export { logger } from './logger';
```
Actually we can only keep `index.ts`

osipov-mit (approved):

PR #1761, opened 2025-01-30
introduce gearexe idea frontend

vraja-nayaka:


ereburg (approved):

nikitayutanov:
Please don't merge until complete review.

idea/gearexe/frontend/tsconfig.app.json
```diff
@@ -0,0 +1,31 @@
+{
```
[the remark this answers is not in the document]
nikitayutanov: I wouldn't do so, because semantically they serve different purposes.

It follows the default template provided by Vite, which helps in noticing changes in case of an upgrade.

idea/gearexe/frontend/src/shared/config/consts.ts
```diff
@@ -0,0 +1,29 @@
+export const PROJECT_ID = import.meta.env
+  .VITE_WALLET_CONNECT_PROJECT_ID as string;
+export const GEAR_API_NODE = import.meta.env.VITE_GEAR_EXE_API_NODE as string;
+export const ETH_CHAIN_ID = 17000; // (0x4268) Holesky
+export const ETH_NODE_ADDRESS = import.meta.env.VITE_ETH_NODE_ADDRESS;
+
+export const DIGIT_RECOGNITION_CONTRACT_ADDRESS = import.meta.env
+  .VITE_DIGIT_RECOGNITION_CONTRACT_ADDRESS as `0x${string}`;
+
+enum AnimationTimeout {
```
nikitayutanov, line 10: I would suggest not to utilize enums, `as const` objects should be sufficient.

[This approach is opinionated](https://www.youtube.com/watch?v=jjMbPt_H3RQ), but [TS docs](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums) suggests we might not need them as well.

idea/gearexe/frontend/src/shared/utils/merge-refs.ts
```diff
@@ -0,0 +1,15 @@
+import type * as React from "react"
+
+export function mergeRefs<T = any>(
```
nikitayutanov, line 3: Not needed for now.

idea/gearexe/frontend/vite.config.ts
```diff
@@ -0,0 +1,11 @@
+import react from '@vitejs/plugin-react';
+import path from 'path';
+import { defineConfig } from 'vite';
+import svgr from 'vite-plugin-svgr';
+
+// https://vite.dev/config/
+export default defineConfig({
+  plugins: [react(), svgr()],
```
nikitayutanov, line 8: Add `vite-plugin-checker`.

nikitayutanov:
Let's stick to same folder and file naming conventions as in Idea - `hyphens-case`, lowercase.
For example: directory `nodeSwitch` - `node-switch`, component `Layout.tsx` - `layout.tsx`.

It helps to mitigate some problems in case sensitive environments.

idea/gearexe/frontend/src/features/api/consts.ts
```diff
@@ -0,0 +1,8 @@
+import { LocalStorage, GEAR_API_NODE } from '@/shared/config';
+
+import { getNodeAddressFromUrl } from './utils';
+
+// ! TODO: move to nodesSwitch
+const INITIAL_ENDPOINT = getNodeAddressFromUrl() || (localStorage[LocalStorage.Node] as string | null) || GEAR_API_NODE;
```
nikitayutanov, line 6: Let's move it directly. No reason to keep `api` feature, I think it's quite broad and should be shared.

idea/gearexe/frontend/src/components/ui/badge/Badge.tsx
```diff
@@ -0,0 +1,19 @@
...
+import styles from "./Badge.module.scss";
+
+type Props = PropsWithChildren & {
+  color?: number | "primary" | "secondary";
+  className?: string;
+};
+
+export const Badge = ({ children, className, color = "primary" }: Props) => {
+  const variant =
+    typeof color === "number" ? `color-${(color % 10) + 1}` : color;
```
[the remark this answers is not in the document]
vraja-nayaka: The idea is that each service name will have a fixed, unchanging color assigned to it.

vraja-nayaka:
> Not sure about `search` though, maybe would make more sense to just have `search-form` component.

The search will encapsulate the logic for sending API requests to determine what the user has entered: a code, a program, a wallet, or something else.

I don't want to keep this logic in app.tsx.

idea/gearexe/frontend/src/shared/config/routes.ts
```diff
@@ -0,0 +1,9 @@
+const routes = {
+  home: "/",
+  programs: "/programs",
```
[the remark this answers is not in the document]
vraja-nayaka: I suggest splitting this by features as the number of entities increases.

Right now, I don't think it makes sense.
We only have five routes, and it's not yet clear which features will use the environment variables.
It's quite possible that many of them will be needed in multiple features.

idea/gearexe/frontend/src/components/layout/Layout.module.scss
```diff
@@ -0,0 +1,10 @@
+.container {
+  max-width: 1440px;
+  margin: 0 auto;
+  padding: 8px 48px 0px 48px;
+  height: calc(100vh - 78px);
```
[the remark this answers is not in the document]
vraja-nayaka: The entire design is optimized so that the content fits within a single screen without the need for main scrolling.
We can discuss this on a call.
I changed it to height: 100vh and set flex: 1 on the child element.
Now, the header height doesn’t matter.

idea/gearexe/frontend/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "gearexe-idea-frontend",
+  "private": true,
+  "version": "0.0.1",
+  "type": "module",
+  "scripts": {
+    "start": "npx vite",
+    "build": "rm -rf dist && npx tsc -b && npx vite build",
```
nikitayutanov, line 8:
```suggestion
    "build": "rm -rf dist && npx vite build",
```

With enabled `vite-plugin-checker` there's no need for `tsc`.

idea/gearexe/frontend/src/components/header/header.module.scss
```diff
@@ -0,0 +1,23 @@
+.container {
+  position: relative;
+  width: 100vw;
+  max-width: 1440px;
+  margin: 0 auto;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 16px 48px;
+  height: 78px;
```
nikitayutanov, line 10: If it's possible I wouldn't specify explicit height at all, otherwise it's too easy to encounter overflow issues on different font sizes or future changes in layout.

idea/gearexe/frontend/src/components/navigation/navigation.module.scss
```diff
@@ -0,0 +1,29 @@
...
+  grid-template-columns: 1fr 1fr;
+  border: 1px solid rgba(41, 41, 41, 1);
+  border-radius: 8px 8px 0 0;
+  overflow: hidden;
+}
+
+.navigation {
+  display: flex;
+  gap: 24px;
+  height: 52px;
```
nikitayutanov, line 12: Same as above, I wouldn't hardcode height.

idea/gearexe/frontend/src/components/ui/badge/badge.module.scss
```diff
@@ -0,0 +1,78 @@
+.container {
+  padding: 1px 12px;
+  border-radius: 2px;
+  font-size: 11px;
+  height: 20px;
```
nikitayutanov, line 5: Same as above, I wouldn't hardcode height.

idea/gearexe/frontend/src/components/ui/syntax-highlighter/syntax-highlighter.tsx
```diff
@@ -0,0 +1,36 @@
+import ReactHighlighter, { SyntaxHighlighterProps } from 'react-syntax-highlighter';
+import { stackoverflowDark } from 'react-syntax-highlighter/dist/esm/styles/hljs';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const Highlighter = ReactHighlighter as any as React.FC<SyntaxHighlighterProps>;
```
nikitayutanov, line 5:
```suggestion
const Highlighter = ReactHighlighter as unknown as React.FC<SyntaxHighlighterProps>;
```
Wouldn't be better like that?

idea/gearexe/frontend/src/components/ui/table/table.module.scss
```diff
@@ -0,0 +1,62 @@
...
+
+  &.lineHeight-md {
+    tbody tr {
+      height: 32px;
+    }
+  }
+
+  &.lineHeight-lg {
+    tbody tr {
+      height: 47px;
```
nikitayutanov, line 15: Same as above, I wouldn't hardcode height.

idea/gearexe/frontend/src/components/ui/expandable-item/expandable-item.module.scss
```diff
@@ -0,0 +1,44 @@
...
+.header {
+  display: flex;
+  gap: 12px;
+  align-items: center;
+  justify-content: space-between;
+
+  &.nested {
+    background: rgba(15, 15, 15, 1);
+    padding: 8px 16px;
+    height: 42px;
```
nikitayutanov, line 18: Can we omit fixed height?

Issue #1687, opened 2024-12-03
fix(wallet-connect): required `@gear-js/ui` dependency


nikitayutanov:
### Problem

![image](https://github.com/user-attachments/assets/7f79c6ba-6e60-4f28-89a3-139fb36ef058)

### Steps

1. Install `@gear-js/wallet-connect` with only `@gear-js/vara-ui`
2. Launch app

PR #1873, opened 2025-04-14
fix(wallet-connect): required @gear-js/ui dependency

vraja-nayaka:
Resolves https://github.com/gear-tech/gear-js/issues/1687


nikitayutanov (changes requested):
I tested it locally, and seems like it's a fix.
However I don't like the idea to store entire modules in a state, it just seems too redundant and complicated.

I think we should investigate if it's possible to be done with top-level await imports, but from a quick glance there are some issues with Vite: https://github.com/vitejs/vite/issues/14102, https://github.com/vitejs/vite/issues/15864, https://github.com/tailwindlabs/tailwindcss/issues/17208

Provided solution is solid and we can figure something out of it, just in a different more straightforward manner.

PR #1909, opened 2025-04-30
fix(gear-idea-faucet): fix creating account and running app.

osipov-mit:
Bump polkadot dependencies across the repo


utils/wallet-connect/package.json
```diff
@@ -27,7 +27,7 @@
     "@gear-js/react-hooks": "^0.16.0",
     "@gear-js/ui": "^0.7.0",
     "@gear-js/vara-ui": "^0.3.1",
-    "@polkadot/react-identicon": "^3.12.2",
```
nikitayutanov, line 30: Don’t modify peer deps apart from ‘@polkadot/api’, to bump dev deps should be enough. Its version shouldn’t be bumped since there is no pin, otherwise there are constant dependency conflicts in apps.

nikitayutanov (approved):

PR #1932, opened 2025-05-21
feat(gear-api): introduce  `Program` class

osipov-mit:


apis/gear/src/TxBuilder.ts
```diff
@@ -0,0 +1,3 @@
+export class TxBuilder {
```
nikitayutanov, line 1: Is it a placeholder for a functionality that will be available later?

osipov-mit: forgot to remove

apis/gear/src/GearProgram.ts
```diff
@@ -0,0 +1,375 @@
...
+    this._isInitialized = true;
+  }
+
+  /**
+   * ## Subscribes to a specific event emitted by the program.
+   * @param action - The name of the event to subscribe to (e.g., 'programExited').
+   * @param cb - The callback function to execute when the event is triggered. Receives the inheritor ID as a parameter.
+   * @returns A function to unsubscribe from the event.
+   */
+  public async on(action: 'programExited', callback: (inheritorId: HexString) => void | Promise<void>) {
```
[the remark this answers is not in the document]
osipov-mit: so far. I think there will be more

nikitayutanov:
I believe it would be good to provide some interface (method?) to send messages in a batch.

This is a common use case in dApps, and Bridge is the most recent example where we've encountered this need. The current implementation of `sails-js` generated lib pretty much ties down our hands in this regard. 

Looks like there's a solid foundation for this with `_submitTx` and its `eventsToBeReturned`. I took a similar approach [here](https://github.com/gear-tech/gear-bridges/blob/b4ec44ccb50c870dbc966a841f9c1c634121c472/frontend/src/features/swap/hooks/vara/use-sign-and-send.ts#L77). 

If with its help we're able to provide some ground to batch transaction sending - something that could later be utilized in `sails-js` - now would be a great time and opportunity to implement it.

osipov-mit:
> I believe it would be good to provide some interface (method?) to send messages in a batch.
> 
> This is a common use case in dApps, and Bridge is the most recent example where we've encountered this need. The current implementation of `sails-js` generated lib pretty much ties down our hands in this regard.
> 
> Looks like there's a solid foundation for this with `_submitTx` and its `eventsToBeReturned`. I took a similar approach [here](https://github.com/gear-tech/gear-bridges/blob/b4ec44ccb50c870dbc966a841f9c1c634121c472/frontend/src/features/swap/hooks/vara/use-sign-and-send.ts#L77).
> 
> If with its help we're able to provide some ground to batch transaction sending - something that could later be utilized in `sails-js` - now would be a great time and opportunity to implement it.

Let's make an issue out of this and I'll get to it a bit later

nikitayutanov (approved):

PR #1934, opened 2025-05-23
chore(hooks, idea/frontend): bump api

nikitayutanov:


idea/gear/frontend/src/features/dns/consts/sails.ts
```diff
@@ -1,86 +1,89 @@
...
-  admin: ActorId;
-  program_id: ActorId;
-  registration_time: string;
-}
+import { GearApi, Program } from '@gear-js/api';
+import { TypeRegistry } from '@polkadot/types';
+import { TransactionBuilder, ActorId } from 'sails-js';
 
-export class Program {
+export class SailsProgram {
```
osipov-mit, line 7: don't we want to have a meaningful name here (like DnsProgram or sth else)?

nikitayutanov: All `dns` related data is in its related feature folder, there's no other Sails program in it, therefore I think default `SailsProgram` is sufficient and descriptive. Like, there's `dns` feature -> it has its Sails program.

osipov-mit (approved):

PR #2206, opened 2025-11-28
Feat(vara-eth-frontend): get codes from indexer

vraja-nayaka:


idea/vara-eth/frontend/src/components/ui/pagination/pagination.tsx
```diff
@@ -0,0 +1,44 @@
...
+    if (currentPage > 1) {
+      onPageChange(currentPage - 1);
+    }
+  };
+
+  const handleNext = () => {
+    if (currentPage < totalPages) {
+      onPageChange(currentPage + 1);
+    }
+  };
```
nikitayutanov:
```suggestion
  const handlePrev = () => {
    onPageChange(prevPage => prevPage - 1);
  };

  const handleNext = () => {
    onPageChange(prevPage => prevPage + 1);
  };
```

Should be updated using updater function. Check guards aren't needed - it's handled by button state in UI.

idea/vara-eth/indexer/src/api.ts
```diff
@@ -31,6 +32,12 @@ async function main() {
 
   const app = express();
 
+  app.use(
+    cors({
+      origin: process.env.FRONTEND_URL || 'http://localhost:3000',
+      credentials: true,
+    }),
+  );
```
osipov-mit:
```suggestion
  app.use(cors());
```

osipov-mit (changes requested):

osipov-mit (approved):

PR #2216, opened 2025-12-02
feat(vara-eth-frontend): add truncated text and ui improvements

vraja-nayaka:


idea/vara-eth/frontend/src/shared/utils/styles/_breakpoints.scss
```diff
@@ -0,0 +1,38 @@
+$xl: 1200px;
+$lg: 992px;
+$md: 768px;
+$sm: 576px;
+
+@function getMaxWidth($width) {
+  @return 'max-width: #{$width - 1}';
+}
```
nikitayutanov: No issue, just a side comment:

I'm not sure whether it's needed to subtract 1 or not - it's up to you. When I first made this decision it was based on the fact that most apps still display desktop layout at `768px` and only start to display mobile one at `767px` (probably because `min-width` or some ranges is used everywhere starting with mobile-first approach).

I would like to investigate further what's the preferred approach and would be good to know your opinion as well.

idea/vara-eth/frontend/src/components/ui/media-query/media-query.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import styles from './media-query.module.scss';
+
+function Desktop({ children }: PropsWithChildren) {
+  return <span className={styles.desktop}>{children}</span>;
+}
+
+function Mobile({ children }: PropsWithChildren) {
+  return <span className={styles.mobile}>{children}</span>;
+}
```
nikitayutanov: Looks like `Desktop` and `Mobile` are not needed anymore.

idea/vara-eth/frontend/src/components/ui/media-query/media-query.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type BreakpointSize = 'Sm' | 'Md' | 'Lg' | 'Xl' | 'Xxl';
```
nikitayutanov:
```suggestion
type BreakpointSize = 'sm' | 'md' | 'lg' | 'xl' | 'xxl';
```

Up to you, but seems easier to write.

PR #2219, opened 2025-12-03
fix(vara-eth-frontend): fix ethereum client init

vraja-nayaka:


nikitayutanov (changes requested):

idea/vara-eth/frontend/src/app/api/use-ethereum-client.ts
```diff
@@ -1,14 +1,30 @@
...
-export function useEthereumClient({ chainId }: { chainId?: number } = {}) {
-  const { data: walletClient } = useWalletClient({ chainId });
-  const publicClient = usePublicClient({ chainId });
+export function useEthereumClient() {
+  const { data: walletClient } = useWalletClient({ chainId: ETH_CHAIN_ID });
+  const publicClient = usePublicClient({ chainId: ETH_CHAIN_ID });
 
-  if (!walletClient || !publicClient) return null;
+  return useQuery({
+    queryKey: ['ethereumClient', ETH_CHAIN_ID, walletClient?.account?.address, publicClient?.chain?.id],
```
nikitayutanov:
```suggestion
    queryKey: ['ethereumClient', walletClient, publicClient],
```

`ETH_CHAIN_ID` is static, no need for it.

Is it possible to pass `walletClient` and `publicClient` as a whole? Both of them can contain some dynamic properties that can change and we will not be aware about it. The one that I encountered by myself - `walletClient` can have a different chain id until it's synced with `publicClient`.

idea/vara-eth/frontend/src/app/api/use-mirror-contract.ts
```diff
@@ -1,19 +1,24 @@
...
 
 const useMirrorContract = (mirrorAddress: Address) => {
   const ethAccount = useAccount();
   const { data: walletClient } = useWalletClient({ chainId: ethAccount.chain?.id });
   const publicClient = usePublicClient({ chainId: ethAccount.chain?.id });
 
-  if (!walletClient || !publicClient) {
-    return { mirrorContract: null, isLoading: true };
-  }
+  return useQuery({
```
nikitayutanov: Same comments as above - `queryKey` should be as broad as possible, `queryFn` should return only awaited result - everything else is an error.

PR #2270, opened 2025-12-24
feat(gear-idea/explorer): add search and timestamp filtering for messages and events 

nikitayutanov:


nikitayutanov:
@osipov-mit, I'm not sure how to interpret `to` parameter as the end of the day in case date without time is provided, let me know please.

idea/gear/explorer/src/services/message.ts
```diff
@@ -122,6 +140,18 @@ export class MessageService {
       qb.andWhere('msg.fn ILIKE :fn', { fn: `%${fn.toLowerCase()}%` });
     }
 
+    if (query) {
+      qb.andWhere('msg.id ILIKE :query', { query: `%${query.toLowerCase()}%` });
```
osipov-mit: What's the use case of such querying?

osipov-mit (approved):

PR #2272, opened 2025-12-25
feat(vara-eth-idea/frontend): migrate from Graphql to REST api

vraja-nayaka:


idea/vara-eth/frontend/src/features/codes/lib/queries.ts
```diff
@@ -1,56 +1,51 @@
...
+  const limit = pageSize;
   const offset = (page - 1) * pageSize;
 
   return useQuery({
-    queryKey: ['allCodes', first, offset],
-    queryFn: () => request(EXPLORER_URL, GET_CODES_QUERY, { first, offset }),
-    select: (data) => data.allCodes,
+    queryKey: ['allCodes', limit, offset],
+    queryFn: async () => {
+      const url = new URL(`${EXPLORER_URL}/codes`);
```
osipov-mit: Don't you want to move this to fetchWithGuard fn since it's the same in all queries except for the endpoint?

idea/vara-eth/frontend/src/features/codes/lib/queries.ts
```diff
@@ -1,56 +1,51 @@
...
+  id: string;
+  status: string;
+  createdAt: string;
+};
+
+type CodesResponse = {
+  data: Code[];
+  total: number;
+  limit: number;
+  offset: number;
```
osipov-mit: I guess it can be a separate type like `ListResponse`. And types like this could extend it

idea/vara-eth/frontend/src/features/codes/lib/queries.ts
vraja-nayaka: I don't think we should move this part into fetchWithGuard. In the queryFn we're only setting the URL endpoint, and that’s the only piece that differs across queries. It feels unnecessary to change fetchWithGuard for this case.

idea/vara-eth/frontend/src/features/codes/lib/queries.ts
```diff
@@ -1,56 +1,47 @@
...
-  }
-`);
-
 export const useGetCodeByIdQuery = (id: string) => {
   return useQuery({
     queryKey: ['codeById', id],
     queryFn: async () => {
-      const result = await request(EXPLORER_URL, GET_CODE_BY_ID_QUERY, { id });
-      return result;
+      const url = new URL(`${EXPLORER_URL}/codes/${id}`);
```
nikitayutanov: Any particular reason to use `new URL` instead of `string`? If there're currently no plans to have search params for urls like that, I believe better to keep things simple.

nikitayutanov (approved):

PR #2284, opened 2026-01-08
feat(vara-eth-idea/frontend): upload idl to local storage

vraja-nayaka:


idea/vara-eth/frontend/src/shared/hooks/use-idl-storage.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const getIdlQueryKey = (codeId?: HexString) => ['idl', codeId];
+
+type UseIdlStorageReturn = {
+  idl: string | null;
+  saveIdl: (idlContent: string) => void;
+  isLoading: boolean;
+};
+
+export const useIdlStorage = (codeId?: HexString): UseIdlStorageReturn => {
```
nikitayutanov: I believe it's overengineered a little bit.

`@tanstack/react-query` is async state manager and there aren't any async operations when we communicate with local storage.

Simple `useState(localStorage.getItem(IDL_LOCAL_STORAGE_KEY));` at the program and code page, while passing `setState` to `UploadIdlButton` `onChange` prop should be sufficient.

And if not, in the future we can take a closer look at `useSyncExternalStore` solutions.

PR #2332, opened 2026-02-04
feat(vara-eth-api): introduce ISigner usage for signing and sending transactions

osipov-mit:


apis/vara-eth/src/signer/errors.ts
```diff
@@ -0,0 +1,21 @@
+export class SigningError extends Error {
+  constructor(
+    message: string,
+    public readonly cause?: Error,
+  ) {
+    super(message);
```
nikitayutanov:
```suggestion
    cause?: Error,
  ) {
    super(message, { cause });
```

I believe native `cause` is supported.

apis/vara-eth/src/eth/mirror.ts
```diff
@@ -13,27 +13,25 @@ import {
...
    * @param address - The address of the Mirror contract
-   * @param walletClient - The wallet client for sending transactions and signing messages
+   * @param signer - The signer for sending transactions and signing messages
    * @param publicClient - The public client for reading contract data
    */
   constructor(
     public readonly address: Address,
-    private _wc: WalletClient<TTransport, TChain, TAccount>,
+    private _signer: ISigner,
     private _pc: PublicClient<TTransport, TChain>,
```
nikitayutanov: Let's make `signer` optional.

Recently I figured out that `WalletClient` is `undefined` if wallet is not connected, but we might still need `MirrorClient` instance to read state.

Related to #2326.

PR #2435, opened 2026-04-14
feat(apis/vara-eth): precalculate blob versioned hash

osipov-mit:


vraja-nayaka (approved):

osipov-mit:
/gemini review the changes one more time

PR #2441, opened 2026-04-16
feat(vara-eth-idea/frontend): create program with abi

vraja-nayaka:
UI: 
{8FE942A2-3085-4DC1-8457-DAD80E1CC2F1}


vraja-nayaka:
replaced with https://github.com/gear-tech/gear-js/pull/2457

PR #2464, opened 2026-05-05
feat(apis/vara-eth): support new approach to signing promises

osipov-mit:
Related to https://github.com/gear-tech/gear/pull/5377

### DIFF
--- a/apis/vara-eth/src/api/injected/promise.ts
+++ b/apis/vara-eth/src/api/injected/promise.ts
@@ -1,3 +1,4 @@
+import { blake2b } from '@noble/hashes/blake2';
 import { keccak_256 } from '@noble/hashes/sha3.js';
 import type { Address, Hash, Hex } from 'viem';
 import { bytesToHex, concatBytes, hexToBytes, recoverMessageAddress } from 'viem';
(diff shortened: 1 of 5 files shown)


apis/vara-eth/src/api/injected/promise.ts
```diff
@@ -90,11 +91,30 @@ export class InjectedTxPromise implements IInjectedTransactionPromise {
...
+
+  private get _replyHashU8a(): Uint8Array {
+    return blake2b(this._replyDataU8a, { dkLen: 32 });
+  }
+
+  public get replyHash(): Hash {
+    return bytesToHex(this._replyHashU8a);
+  }
+
+  public get compactPromise() {
```
kuzminxyz: Please make it private if you can. `CompactPromise` is only vara.eth internal type, so it shouldn't be provided to user in any possible way.

ecol-master (approved):

apis/vara-eth/src/api/injected/promise.ts
osipov-mit: I don't even need it, forgot to remove

vraja-nayaka (approved):

PR #2478, opened 2026-05-12
feat(idea/vara-eth/backend): prepare for indexing injected transactions

osipov-mit:


biome.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.4.15/schema.json",
```
vraja-nayaka, line 2: why we downgrade it?

vraja-nayaka (approved):

PR #2483, opened 2026-05-15
feat(apis/vara-eth): 0.5.0-rc.0 — helpers, streams, typed errors

ukint-vs:
Bumps `@vara-eth/api` 0.4.0 → **0.5.0-rc.0**. Lands the lib-side primitives a full-featured wallet — dApp or CLI — needs, without breaking the adapter-shaped contract. The lib still doesn't hold keys; consumers pass an `ITransactionSigner` and pick their rail.

### Two signer rails, one API

```ts
// dApp (browser): MetaMask / WalletConnect / any EIP-1193 wallet
import { walletClientToSigner } from '@vara-eth/api';
const signer = walletClientToSigner(walletClient);

// CLI / agent: local key, V3 keystore handled at the wallet layer (out of this PR)
import { LocalSigner } from '@vara-eth/api';
const signer = new LocalSigner(privateKey, publicClient);

// Same call from here on, either rail:
await api.programs.deploy(wasmBytes, { signer, executableBalance: 10n ** 18n });
await api.programs.sendAndWait(mirror, payload, { signer, via: 'eth' });
```

The lib never sees a private key on the dApp rail; on the CLI rail the key custody choice lives entirely in the consumer (keystore, mnemonic, Ledger — all out of scope here by design).

### Helpers that collapse multi-step ceremonies

- `api.programs.deploy(code, opts)` — WVARA permit → `requestCodeValidation` → bounded `CodeGotValidated` wait → fresh executable-balance permit (if requested) → `createProgram*` variant pick → result.
- `api.programs.sendAndWait(mirror, payload, opts)` — on-chain `Mirror.sendMessage` (default `via: 'eth'`) or off-chain injected via `injected_sendTransactionAndWatch`. Uniform `{messageId, reply, txHash, validator?}` result.
- `api.fees.estimate(op)` — viem-backed gas + WVARA fee preview before submit.
- `api.stream.{programEvents,routerEvents,blocks}(handlers, opts?)` — typed discriminated-union event streams over viem's `watchContractEvent` / `watchBlocks`.
- `extractSailsIdl{,OrThrow}` — pure WASM custom-section parser, tolerates both `sails_idl` and `sails-idl` names.

### Typed errors at public boundaries

`VaraEthError` base + named subclasses for everything callers might want to branch on: `ViemForkRequiredError`, `InjectedTxStaleError`, `PromiseTimeoutError`, `PromiseSignatureInvalidError`, `PermitExpiredError`, `BlobUnderpricedError`, `CodeValidationTimeoutError`, `NoSailsIdlError`, `RpcConnectionError`, `ChainIdMismatchError`. Internal throw sites untouched — wallet-side errors (`WalletLockedError`, …) stay in the CLI by design.

### Phase 0 — gates + packaging
- `scripts/poc-wallet.ts` runs the full ceremony against `ethexe run --dev` (`yarn poc:ethexe`). `yarn typecheck:poc` runs in CI on every PR so the script can't silently rot when lib types change.
- `assertViemFork()` runtime check at `RouterClient.requestCodeValidation*` entry points only. EIP-7594 blob upload requires `@vara-eth/viem@2.48.11`; upstream viem trips `ViemForkRequiredError` with a remediation message. Read-only consumers never hit the check.
- **JS-side signing golden fixture** pins preimage byte layout, keccak256 hash, blake2b messageId, and the deterministic ECDSA signature — catches EIP-191 / raw-bytes / recovery-byte drift before it ships.
- **CJS packaging fix.** `lib/cjs/` now ships `{"type":"commonjs"}` so Node doesn't read its `require()`-based files as ESM. Subpath `require` exports point at real built files. Top-level `main` + `types` added.

### Phase 1 — helpers + typed errors + Sails IDL
See "What this gives consumers" above.

### Phase 2 — event streams
Thin discriminated-union wrappers over viem watchers — reconnection delegated to viem.

## Review history

Two independent reviews on the first push (`d76776ae`) — Claude eng-review + `/codex review` — found 6 disjoint blockers. ~10% agreement rate, because there were almost zero unit tests for the new helpers and each reviewer caught different classes of bug.

**3 P1 regressions** (broke on first call, found by codex):
- `sendAndWait(via:'eth')` was missing `tx.send()` between `sendMessage` and `setupReplyListener` — the listener throws immediately without it.
- `LocalSigner` cast `publicClient.transport` to `Transport`, but viem treats `transport` as a factory. Replaced with `custom({ request: publicClient.request })`.
- `viem-fork.ts` did a top-level `import { createRequire } from 'node:module'`, which poisons browser bundles since the module is reachable from the root `@vara-eth/api` index. Moved the import behind a Node-runtime guard with `Function('return typeof require !== "undefined" ? require : null')()` for opaque indirection.

**3 deploy-ceremony reliability bugs** (eng-review):
- `deployProgram` had no timeout on `waitForCodeGotValidated()` — would hang forever if validators stalled. Now bounded by `codeValidationTimeoutMs` (default 120 s); on expiry throws `CodeValidationTimeoutError` carrying `codeId` + `txHash` so the caller can resume the second half out-of-band.
- Executable-balance permit was signed in parallel with the validator wait, sharing one absolute deadline with the code-fee permit. If the wait exceeded 5 minutes (default permit window), `createProgramWithExecutableBalance` would silently revert. Now signed AFTER `CodeGotValidated` resolves, with a fresh `now`-based deadline.
- `InjectedTxStaleError` was declared but never thrown. Now thrown from `InjectedTx.setReferenceBlock(supplied)` when the supplied block is outside the 32-block validity window. Transient RPC failures fall through silently; programmer errors (`TypeError`/`RangeError`) are re-thrown.

Each bug now has a fence test in `test/unit/`. CHANGELOG reconciled (removed stale `MessageRevertedError` mention, added `CodeValidationTimeoutError` + `codeValidationTimeoutMs`). Post-review `/simplify` pass dedup'd the `Promise.race`-with-timeout pattern into `src/util/promise.ts#withTimeout` and parallelized the staleness pre-check.

## Test plan

- [ ] CI green
- [ ] `yarn workspace @vara-eth/api typecheck:poc` passes
- [ ] `yarn workspace @vara-eth/api test:unit` — **50 tests pass** (40 pre-existing + 6 fence tests for the blockers + 4 across IDL/local-signer fixtures)
- [ ] `yarn workspace @vara-eth/api build` — both ESM and CJS outputs produced
- [ ] CommonJS smoke: \`node -e "const a = require('@vara-eth/api'); const v = require('viem'); new a.LocalSigner('0x' + 'ab'.repeat(32), v.createPublicClient({transport: v.http('http://x')}));"\` — does not throw
- [ ] Manual: `yarn workspace @vara-eth/api poc:ethexe` against `ethexe run --dev`

## Notes for reviewers

- New public surface in 0.5.0-rc.0: `LocalSigner`, `privateKeyToLocalSigner`, `WalletClientAdapter`, `walletClientToSigner`, `ProgramsNamespace`, `FeesNamespace`, `StreamNamespace`, all typed errors, `extractSailsIdl{,OrThrow}`, `watchProgramEvents`/`watchRouterEvents`/`watchBlocks`, `buildEventMeta`.
- Validator targeting in `sendAndWait` defaults to broadcast (zero-address recipient); pool round-robin is opt-in only — slot-based selection is heuristic and worse than broadcast under clock skew / reorgs / validator-set transitions.
- Out of scope by design: keyring, V3 keystore, mnemonic, HD derivation, Ledger, Sails typed-client codegen, operator/validator mode, cross-chain "deploy to both" workflow. These live in (or arrive with) the wallet CLI consumer, not this lib.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ukint-vs:
## Fix-up pushed — addresses 6 blockers from `/codex review` + internal eng review

**Codex P1s (functional regressions, every helper threw on first call):**
- 1.1 `api.programs.sendAndWait({via:'eth'})` now calls `tx.send()` before `setupReplyListener()`. The default on-chain path was throwing `"No transaction hash available. Call send() first."` every time.
- 1.2 `LocalSigner` routes through `custom({ request: publicClient.request })` instead of casting the already-constructed transport. The constructor no longer throws.
- 1.3 `@vara-eth/api` is now importable from browser bundles. Dropped the top-level `import { createRequire } from 'node:module'` in `src/util/viem-fork.ts` that crashed esbuild/webpack at bundle resolution. **This was a hard blocker for every dApp using MetaMask via the existing `walletClientToSigner` path.**

**Deploy-ceremony fixes (Claude eng-review):**
- 2.1 `deployProgram` now bounds the `CodeGotValidated` wait with `codeValidationTimeoutMs` (default 120s). On expiry throws new `CodeValidationTimeoutError` carrying `{ codeId, txHash, timeoutMs }` so callers can resume out-of-band.
- 2.2 Executable-balance permit is signed AFTER `CodeGotValidated` resolves with a fresh `now`-based deadline. Previously shared the code-fee permit's 5-min deadline → silent revert + burned code-validation fees on long validator waits.
- 2.3 `InjectedTx.setReferenceBlock(suppliedHash)` pre-checks against chain head and throws `InjectedTxStaleError` for blocks outside the 32-block validity window. Transient RPC failure falls through.

**Tests:** 50 passing (40 prior + 10 fence). Each fix has a unit test that would have caught the bug.

**CHANGELOG:** reconciled — dropped `MessageRevertedError` + `StreamHandlers.onReconnect` mentions (never shipped); added `CodeValidationTimeoutError` + `codeValidationTimeoutMs`.

Ready for re-review.

PR #2556, opened 2026-06-24
feat(idea/frontend): add basic idl v2 support

vraja-nayaka:


utils/gear-hooks/src/hooks/sails/parse-idl.ts
```diff
@@ -0,0 +1,28 @@
...
+
+type ParsedSails = Sails | SailsProgram;
+
+const isIdlV2 = (idl: string) => /!@sails:/.test(idl);
+
+type SailsParser = (idl: string) => ParsedSails;
+
+const createSailsParser = async (): Promise<SailsParser> => {
+  const v1Parser = new V1Parser();
+  const v2Parser = new V2Parser();
```
osipov-mit: It seems as not the best idea to always create both parsers. It works fine for Gear/Vara.Eth Idea but for smaller applications working with a single contract it can be an overhead because of 2 WASM modules is loaded and initialized here. But, idk where the library is used actually

osipov-mit (approved):

utils/gear-hooks/src/hooks/sails/parse-idl.ts
vraja-nayaka: I've switched to lazy initialization: each WASM parser is now loaded only on the first parse of the matching IDL version (!@sails: → v2, otherwise v1). For a single-contract app that means only one WASM module is initialized.

